PR #5, opened 2020-02-15
Show graphical characters instead of alphabet on ovmenu

CazYokoyama:
ovmenu showed alphaet characters as surrounding frame. Replace them by graphical characters. 

Signed-off-by: Caz Yokoyama <caz@caztech.com>


kedder:
:+1:

Issue #2, opened 2020-02-01
Porting to yocto/warrior: making XCSoar run


kedder:
One of the major issues with XCSoar on current `warrior` branch is inability to rotate the screen to match the device orientation. XCSoar runs under linux framebuffer device and doesn't react to its `DisplayOrientation` configuration option, always running in landscape mode.

Possible solutions I can see: 

1. Switch back to old kernel-3.4 with proprietary mali drivers (known-to-work configuration)
2. Run xcsoar under xorg where it should be (is it?) possible to rotate the screen on Xorg level.

Is there any other option here?

kedder:
So far I have tried option 1:

Kernel 3.4 compiles and boots, but not completely - fails on systemd startup. Apparently systemd requires newer kernel (Linux kernel >= 3.13, according to https://github.com/systemd/systemd/blob/master/README) and will not start. This option would require either downgrading systemd (to some pre 2018 version) or patching it, or switching to systemv init.

None of these are simple/future proof, so I didn't try to continue with it. If completely stuck, we might try to revisit it.

kedder:
For option 2: I was trying to play with new "lima" driver and get it working under Xorg. What I was able to do so far is:

1. Make lima kernel module compile and load
2. Build new mesa (19.1.6) with lima driver
3. Start Xorg and start xcsoar on it.

So far xcsoar runs unsatisfactory slow under this configuration and I'm not sure any accelelartion is coming from lima (probably not). 

On the positive side: `eglgears` do spin under X.

kedder:
Another way to run XCSoar would be through DRM. XCSoar needs to be compiled with these flags:
```
ENABLE_MESA_KMS=y GEOTIFF=n OPENGL=y EGL=y
```

This requires at least kernel with lima driver and mesa with lima driver. Unfortunately so far I couldn't make it work (both kernel and mesa lima drivers present). XCSoar fails to start with this message:

```
Using DRI device /dev/dri/card0 (use environment variable DRI_DEVICE to override)
drmModeGetResources() failed
```

It seems like we need to make `kmscube` running first to make sure infrastructure actually works. It fails with similar error right now: 
```
drmModeGetResources failed: Operation not supported
failed to initialize legacy DRM
```

Looks like there is a thread dealing with similar issue here: https://gitlab.freedesktop.org/lima/linux/issues/22

kedder:
[...]

PR #7, opened 2020-02-21
Enable lima module

kedder:
Enable lima module for both kernel and mesa. Make sure lima loads after sun4i drm for correct device naming in `/dev/dri`.

This is a pre-requisite for #2. Make sure to merge #6 before this one.


kedder:
Machine configs obviously need deduplication, but let's do that in separate PR.

PR #9, opened 2020-02-23
Add XCSoar-7.0-preview12 to the image

kedder:
Introduce XCSoar-7.0-preview12 recipe and use it as default `xcsoar` package for the `openvario-image-testing`. This is the latest known release of xcsoar that supports screen rotation.

Bonus: include default map to the image and make scripts that configure xcsoar actually work (convert from DOS line endsings to native UNIX)

Prerequisite for #2.


kedder:
Makes sense, let me try.

kedder:
I've introduced `openvario-image` target and included stable versions of variod and sensord there as well, Please take a look.

PR #13, opened 2020-03-03
Script mods

iglesiasmarianod:
Here are the mods to the scripts and script files for you to check please.

Thanks.


recipes-apps/ovmenu-ng/files/openvario-57-lvds/ovmenu-ng.sh
```diff
@@ -318,6 +321,14 @@ function download_igc_files() {
 	dialog --backtitle "OpenVario" --title "Result" --tailbox /tmp/tail.$$ 30 50
 }
 
+# Delete only *.igc files
+function delete_igc_files() {
+	echo "Deleting IGC files ..." > /tmp/tail.$$
+	/usr/bin/delete-igc.sh >> /tmp/tail.$$ &
+	dialog --backtitle "OpenVario" --title "Result" --tailbox /tmp/tail.$$ 30 50
```
kedder: I think destructive operation like this needs at least a confirmation dialog. You don't want to accidentally delete your files if you accidentally miss Download_IGC action.

iglesiasmarianod: Hi kedder and thanks!
I'll see how to add a confirmation dialog just in case.

mihu-ov: At least in a club environment easily accessible destructive operations could be an issue. Do we really need a "Delete IGC" option in ovmenu? Disk space on the µSD card is no issue so this is probably more about not having to download all igc files again and again with "Download IGC"? 
What about a "Download newest IGC" option instead? File dates are incorrect on the OV (no battery buffered RTC) but the date is in the IGC file name. Or maybe someday we will be able to select files from a list and upload / download them with a button press.

kedder: I'm kind of with @mihu-ov on this. In what circumstances would you want to delete all your flight logs?

mihu-ov: Can we merge the "Done" message even if we have no final decision on the "Delete IGC" option? I would consider that a small but very nice improvement.

iglesiasmarianod: Hi guys. I've installed two OV in our club´s sailplanes and the question about how to delete IGC files was one of the most often asked questions I had. I agree a destructive method should have a confirmation. I also believe the option to delete must be there. 
All the XCSoar logs are not IGC so they are useless for contests. When you try to download a log from a FLARM to openvario (because FLARM is not accessible to download directly) you have to wait a full IGC download to get the file you want. The more the plane flies, the more you have to wait. I've added the option and what most of our guys do is delete all IGC, fly and then download the FLARM log to send to contest scoring. You are not deleting the contest valid FLARM or logger IGC file so in the worst case scenario no harm is done.
I also agree that a kinda "File Commander" to select which files to download or to delete would be the best option. I just saw this as a temporary patch for the time being.
Sorry if I posted this as a pull request. It's my first time using GitHub as part of a group and I did not know how to share these ideas. If you prefer redmine or XCSoar forum to share please let me know.

Thanks for your time and comments!!
BR

PD: I can create a Branch with only "DONE!" changes so we can merge.

kedder: > When trying to download a log from FLARM to openvario (because FLARM is not fisically accessible for download) you have to wait a full IGC download to get the file you want. The more the plane flies, the more you have to wait.

@iglesiasmarianod, but isn't the root of the problem is that OpenVario downloads *all* the logs all the time, even though most of the time you need only the last one (or all logs for the last day, to generalize). If we would have this option, nobody would want to delete all files, right?

mihu-ov: What if we don´t delete but move the old logs from ~/.xcsoar/logs to some other directory as a compromise? We had several club members who downloaded their flight from the logger to the OV but not from OV to USB drive and realized it only a few days later. Their old logs could still be copied to USB in a shell and not be lost.

iglesiasmarianod: @kedder you are right! I couldn't figure out yet how to "know" what the last file is without RTC.
I started wondering about filling a Menu dialog with filenames inside the log folder.
Something like this (found in stackoverflow):

filter() 
{
    NUM=0
    ls | egrep 'somefilter' | while read i ; do
        echo $'\t'$NUM$'\t'$i
        NUM=$(($NUM+1))
    done
}

LINES=$(filter)

IFS=$'\n\t'

dialog --backtitle "Linux Shell Script" --title "Main\
Menu" --menu "Move using [UP] [DOWN],[Enter] to\
Select" 15 50 3 $LINES 2>/tmp/menuitem.$$

unset IFS

And to retrieve the filename to download this

ITEM=$(cat /tmp/menuitem.$$)
FILE=$(echo "$LINES" | sed -n 's/\t'$ITEM'\t//p')

Do you think it might worth a try?

kedder: @iglesiasmarianod, this would make sense to me. This would be very similar to how you download logs from Flarm - just pick files you want to download. 

Just make sure newest files come first in that list (sort by filename in reverse order).

iglesiasmarianod:
Hi guys, now you can select wich file to download.
One at a time.

BR

recipes-apps/ovmenu-ng-skripts/ovmenu-ng-skripts_0.1.bb
```diff
@@ -19,7 +19,8 @@ SRC_URI =      "\
 	file://download-igc.sh \
 	file://upload-all.sh \
 	file://upload-xcsoar.sh \
-	file://download-xcsoar.sh \
+	file://download-all.sh \
```
kedder: Why the rename?

The recipe in the current form fails to build, because `download-xcsoar.sh` is still referenced in this file below.

recipes-apps/ovmenu-ng/files/openvario-57-lvds/ovmenu-ng.sh
```diff
@@ -171,6 +173,7 @@ function submenu_xcsoar_lang() {
 		 pl_PL.UTF-8 "Poland" \
 		 cs_CZ.UTF-8 "Czech" \
 	 	 sk_SK.UTF-8 "Slowak" \
+	 	 es_ES.UTF-8 "Espanol" \
```
kedder: Let's make the additional language a separate PR. I think it can be included independently from the IGC download stuff.

Note, that right now there are 4 copies of this menu scripts for all the variants of openvario (for whatever reason). You probably want to change all of them.

recipes-apps/ovmenu-ng/files/openvario-57-lvds/ovmenu-ng.sh
```diff
@@ -50,6 +50,7 @@ function submenu_file() {
 	--begin 3 4 \
 	--menu "You can use the UP/DOWN arrow keys" 15 50 4 \
 	Download_IGC   "Download XCSoar IGC files to USB" \
+	Delete_IGC   "Deletes saved IGC files" \
```
kedder: Do you still want to include igc wiping with this feature? To my understanding, file selection should remove the need for it.

recipes-apps/ovmenu-ng-skripts/files/download-igc.sh
```diff
@@ -1,16 +1,39 @@
...
+unset IFS
+
+ITEM=$(cat /tmp/menuitem.tmp)
+FILE=$(echo "$LINES" | sed -n 's/\t'$ITEM'\t//p')
+
+#echo $FILE
+
+cp $IGC_PATH/$FILE $USB_PATH/$FILE
+umount	/dev/sda1
+dialog --backtitle "Openvario" --title "Download IGC" --msgbox "Done!!" 6 20
```
kedder: This "Done" message in the end has one bad consequence: it hides errors from the copy commands. E.g. if your flash drive is not mounted, no files are copied, but you see the "Done" message, that can mislead you into thinking that you've got your igc file.

I think it would be better to just echo "Done" and preserve the log. Make sure you test failure casees too.

recipes-apps/ovmenu-ng-skripts/ovmenu-ng-skripts_0.1.bb
iglesiasmarianod: This was renamed because the script was misnamed since the previous bild system. The feature never worked because it looked for "donwload.xcsoar" while the script was named "download-all"

recipes-apps/ovmenu-ng/files/openvario-57-lvds/ovmenu-ng.sh
iglesiasmarianod: OK, no problem with that. I'll remove it.

recipes-apps/ovmenu-ng/files/openvario-57-lvds/ovmenu-ng.sh
iglesiasmarianod: Let's remove it. With file selection this is not as important now.

recipes-apps/ovmenu-ng-skripts/files/download-igc.sh
iglesiasmarianod: You are right. Saw this after testing.

iglesiasmarianod:
I used a previously built image I had running. Did not try to build it with Warrior yet.
Let me build and see if there are any more mistakes in the buildchain.

iglesiasmarianod:
@kedder Implemented your comments and built the image.
Works out of the box. Would you please test it and let me know if it builds and works in your system?

Thanks!
Regards!

recipes-apps/ovmenu-ng-skripts/files/update-maps.sh
```diff
@@ -22,3 +22,4 @@ fi
 
 
 umount /dev/sda1
+echo "Done !!"
```
kedder: Please make it one exclamation mark and remove the space before it, like this: `Done!`. It is considered an error to put space before punctuation in English (and all the other languages I know for that matter). 

Here and in other places

kedder:
@iglesiasmarianod also, consider rebasing your branch on top of upstream `warrior`. There are too many irrelevant commits in it (like "Merge pull request" ones for example).

iglesiasmarianod:
Think I rebased... (Never done it before).
If it is OK then it's ready to merge with base branch.
Try not to commit so much next time.

linuxianer99:
There are still 21 commits to merge.
Please rebase (https://help.github.com/en/github/using-git/about-git-rebase) the sources so the commits are reduced. 
It is very confusing what has changed in the sources ....

iglesiasmarianod:
@linuxianer99 Hi Timo. Think now I did rebase everything in a single commit.
Please tell me if it is OK.
Thanks for the help and the link.

kedder:
@iglesiasmarianod just look at Commits tab in this PR. I'd expect 1-3 commits for feature like this, and no merge commits for sure. You can squash commits with interactive rebase. 

Typically you add "upstream" remote and rebase on top of it:

```
git remote add upstream https://github.com/Openvario/meta-openvario.git
git fetch upstream
git rebase upstream/warrior --interactive 
# Make sure everything looks good after rebase
git push --force-with-lease
```

iglesiasmarianod:
Guys, messed up.
I'll close the request and let me start clean.
When I have a clean branch I'll request anothe pull.

PR #15, opened 2020-03-08
Enable USB drive automount

### LINKED ISSUE #12: Menu not finding USB drive
Hi guys, I've fixed a misnomer in the Menu scripts to allow file download from OV to USB and added an option to delete all igc stored files.
Also added a Spanish language option to the menu.
Works OK with previous OV buildsystem, now it seems it does not mount USB drives.
If you'd like to have a look to the scripts please, I issued a pull request from my branch.
Let me see if I can figure out why USB seems not to mount.

Regards.

kedder:
This includes autofs with respective kernel module to support USB stick automounting.

Fixes #12


kedder:
Umm, this is not yet ready to merge. Autofs deamon doesn't really boot.

kedder:
Apparently autofs does actually boot, but only after `systemd-networkd-wait-online.service` has started. And that service won't start if you are not online and fail after 2 minute timeout. Only then autofs is allowed to start. I suspect the dependency is there because autofs can also mount network drives. We don't care about network drives though, so it is not really optimal for Openvario...

kedder:
Fixed the issue by removing network dependencies from systemd `autofs.service` descriptor file.

Ready to merge.

mihu-ov:
@kedder: Just tested on the image you provided in https://github.com/Openvario/meta-openvario/issues/18#issuecomment-608081139
Download to USB does not work, the ovmenu scripts expect the USB drive as  /usb/usbstick

```
root@openvario-7-PQ070:~# mount
/dev/mmcblk0p2 on / type ext4 (rw,relatime)
devtmpfs on /dev type devtmpfs (rw,relatime,size=463580k,nr_inodes=115895,mode=755)
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
proc on /proc type proc (rw,relatime)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)
devpts on /dev/pts type devpts (rw,relatime,gid=5,mode=620,ptmxmode=000)
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
cgroup2 on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate)
cgroup on /sys/fs/cgroup/systemd type cgroup (rw,nosuid,nodev,noexec,relatime,xattr,name=systemd)
debugfs on /sys/kernel/debug type debugfs (rw,relatime)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev)
tmpfs on /var/volatile type tmpfs (rw,relatime)
/etc/auto.master.d/auto.usb on /usb type autofs (rw,relatime,fd=5,pgrp=177,timeout=3,minproto=5,maxproto=5,indirect)
root@openvario-7-PQ070:~# /etc/auto.master.d/auto.usb on /usb type autofs (rw,relatime,fd=5,pgrp=177,timeout=3,minproto=5,maxproto=5,indirect)
```

kedder:
@mihu-ov hm, yeah, tried it myself and something is not right, will take a look tomorrow

kedder:
@mihu-ov oh, I was inserting usb stick upside down :)

No, it works for me. In order for automount to work, you need to cd into that dir:

```
root@openvario-57-lvds:~# cd /usb/usbstick
[ 1505.910911] FAT-fs (sda1): Volume was not properly unmounted. Some data may be corrupt. Please run fsck.
root@openvario-57-lvds:/usb/usbstick# ls
KAA_20191231.pdf  openvario
```

mihu-ov:
OK, with cd /usb/usbstick and cd ~ added to /usr/bin/download-igc.sh IGC download works.

#!/bin/bash

USB_PATH="/usb/usbstick/openvario/igc"
IGC_PATH="/home/root/.xcsoar/logs"

cd /usb/usbstick

mkdir -p $USB_PATH

if [ -z "$(ls $IGC_PATH/*.igc 2>/dev/null)" ]; then
        echo "No files found !!!"
else
        for igcfile in $(find $IGC_PATH -name '*.igc'); do
                echo $igcfile
                cp $igcfile $USB_PATH
        done
fi

cd ~

umount /dev/sda1``

kedder:
Hm, cd was just an example, automount will mount the drive on any access to that `/usb/usbstick` directory, e.g. `ls /usb/usbstick/` or `cp file /usb/usbstick/`. So the script should work as it is.

@mihu-ov when it doesn't work, is `/dev/sda1` device present on your device?

PR #22, opened 2020-03-11
Support Texim 7" screens

### LINKED ISSUE #16: Texim 7" display doesn't work
Users are reporting that 7" Texim display doesn't work with the recent (20067) image.

>  Here's a video of non-working image: https://drive.google.com/open?id=1qQbzZW6G60sZAgtfYW7ZHiwC8yZA5LGY

Apparently 7" Texim has different resolution (800x480) than PixelQi one (1024x600). So devicetree files we provide are obviously wrong. U-boot configuration is also wrong. 

We need to pick the right timings from old fex file.

See `recipes-kernel/linux/linux-mainline/openvario-7-CH070.dts`, `recipes-bsp/u-boot/files/openvario-7-CH070/openvario_defconfig`.

kedder:
Timings for Texim 7" was copied from PQ display and wasn't right. I took the new settings from old .fex file and adjusted boot images for Texim resolution (800x480).

Fixes #16 

@linuxianer99, please make sure we provide images for `7-CH070` too.


kedder:
The test by 7" Texim user was positive, so timings seems to be right: https://forum.xcsoar.org/viewtopic.php?f=30&t=3642&p=20725#p20721

PR #33, opened 2020-03-15
Recipe for Syncthing

### LINKED ISSUE #31: Idea: file synchronization
Everybody knows dropbox and it is great for file synchronization across devices. However it has several drawbacks:

* It is a commercial project, and will probably not allow us to redistribute its binaries
* These days they only allow 3 synchronized devices per account on a free plan
* Only allow to sync one folder: `~/Dropbox`

There's a better alternative though: [syncthing](https://syncthing.net/), which is a peer-to-peer file synchronization program.

* It is fully open-source
* No limitation on number of devices
* Has clients for android, linux, windows and macos
* Works without internet (local network is enough)
* It allows to synchronize any number of folders on your system

By a lucky coincidence, XCSoar runs on android phones too. If we share `~/.xcsoar` directory between openvario and the phone, this effectively would allow us to implement this workflow:

* While at home, you configure your xcsoar on your phone the way you like it - pick navboxes, screens, settings, etc; download maps, airspaces, waypoints.
* You arrive at the airfield, go to the briefing and enter today's task to your phone
* You turn on your Openvario, tether internet from your phone and all your configuration and current task gets transferred to Openvario (in a matter of seconds).
* You just launch xcsoar and fly today's task
* After you land, your flight log automatically gets synced back to your phone (and/or a folder in a club' computer, if you want), so you can just go home and analyze it, without dealing with usb sticks.

I'm not aware of such a feature for any of commercial flight computers, so we might even get some envy from their proud owners :)

And I think it should not be super difficult to get it on the openvario image. What do you think?

kedder:
This adds a recipe for [syncthing](https://syncthing.net/) - file synchronization program. This does not include syncthing into the image, but adds a package to opkg repository so users can install it and play with it. 

It is just a stock installation, without any Openvario-specific features. We can experiment with it and figure out the best way to use it.

After installation, syncthing will start its UI on port 8384, it should be accessible under url `http://<openvarioip>:8384`

Fixes #31.

### DIFF
--- a/recipes-apps/syncthing/files/0001-Remove-root-user-warning.patch
+++ b/recipes-apps/syncthing/files/0001-Remove-root-user-warning.patch
@@ -1,0 +1,18 @@
+diff --git a/lib/syncthing/syncthing.go b/lib/syncthing/syncthing.go
+index 99fc23132..77f2bced6 100644
+--- a/lib/syncthing/syncthing.go
++++ b/lib/syncthing/syncthing.go
+@@ -336,10 +336,6 @@ func (a *App) startup() error {
+ 		}
+ 	}
+ 
+-	if isSuperUser() {
+-		l.Warnln("Syncthing should not run as a privileged or system user. Please consider using a normal user account.")
+-	}
+-
+ 	a.evLogger.Log(events.StartupComplete, map[string]string{
+ 		"myID": a.myID.String(),
+ 	})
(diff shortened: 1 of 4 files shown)


kedder:
Not for merge yet - syncthing version is not set up correctly yet

kedder:
Should be good to merge now.

PR #43, opened 2020-03-30
Enable backlight brightness control

kedder:
Allow to change LCD backlight brightness with standard linux tools. Backlight can be controlled by writing to `/sys/class/backlight/lcd/brightness` file. Accepted values are in the range between 0 (backlight off) and 10 (full brightness). For example:

```
echo 8 > /sys/class/backlight/lcd/brightness
```

This also makes screen backlight to turn off when device is switched off in openvario menu.

### DIFF
--- a/recipes-kernel/linux/linux-mainline/openvario-common.dts
+++ b/recipes-kernel/linux/linux-mainline/openvario-common.dts
@@ -227,22 +227,22 @@ cubiets@5c {
 		function = "lvds0";
 	};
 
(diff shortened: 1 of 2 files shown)


linuxianer99:
@kedder is this the same PWM pin for all Openvario variants ??

kedder:
According to `recipes-bsp/u-boot/files/*/openvario_defconfig` PWMs are all on the same pin - `PB2`.

PR #45, opened 2020-04-02
Enable pulseaudio-based variod on testing image

### LINKED ISSUE #18: Vario sound is broken
Vario sound is reported to be broken by several users:

> variod sound output broken

> The audio was odd.  It may have been that things were just incorrectly configured, but my original image, the audio seems very normal compared to a vario.  This  seemed like it was modulated with a low frequency square wave.  I.e. it would just cut in and out constantly.

This gives an idea how it sounds like: https://drive.google.com/file/d/1qQbzZW6G60sZAgtfYW7ZHiwC8yZA5LGY/view

kedder:
This adds pulseaudio on testing image and uses pulseaudio-enabled variod. Switching to pulseaudio-server based sound will also allow to mix XCSoar own sounds with vario.

Requires https://github.com/Openvario/variod/pull/1.

This is supposed to fix #18.

### DIFF
--- a/recipes-apps/variod/files/variod.service
+++ b/recipes-apps/variod/files/variod.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=Variod Daemon for Openvario
 Before=sensord.target
+Requires=pulseaudio.service
+After=pulseaudio.service
 
 [Service]
 Type=forking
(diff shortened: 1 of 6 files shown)


linuxianer99:
OK. so as variod seems to work, i will merge ??

@kedder : Ok ?

kedder:
@linuxianer99 there is one known issue with this - sometimes variod doesn't start. Apparently because some kind of race condition with pulseaudio server. If there's no hurry, let's wait a bit until I fugure it out (expect on Sunday or Monday).

linuxianer99:
ok. fine.
Maybe me build process is also finished, so we can have a huge "test release" :-)

kedder:
Should be good to merge. The problem I mentioned is fixed in https://github.com/Openvario/variod/pull/2

PR #47, opened 2020-04-06
support all leagal file names and keep track of changes

bomilkar:
A simple "cp $downloadfile $USB_PATH" fails with blanks in file names (and other nasty chars).
Use tar instead which can handle all legal file names and has 2 side effects:
1. keep multiple versions on the USB-stick
2. compress on the fly

### DIFF
--- a/recipes-apps/ovmenu-ng-skripts/files/download-xcsoar.sh
+++ b/recipes-apps/ovmenu-ng-skripts/files/download-xcsoar.sh
@@ -2,7 +2,7 @@
 
 USB_PATH="/usb/usbstick/openvario/download/xcsoar"
 DOWNLOAD_PATH="/home/root/.xcsoar"
-SAVE_FILE="${USB_PATH}/xcs-`date +'%Y-%m-%d'`.tgz"
+
 if [ ! -d "$USB_PATH" ]; then
 	mkdir "$USB_PATH"
 fi
@@ -10,11 +10,9 @@ fi
 if [ -z "$(ls $DOWNLOAD_PATH/* 2>/dev/null)" ]; then
         echo "No files found !!!"
 else
-	cd ${DOWNLOAD_PATH}
-	tar czf "${SAVE_FILE}" .
+        cp -rv "$SDOWNLOAD_PATH" "$USB_PATH"
 fi
 
-ls -l "${SAVE_FILE}"
 echo "Umount Stick ..."
 umount /dev/sda1
 echo "Done !!"
--- a/recipes-apps/ovmenu-ng-skripts/files/download-xcsoar.sh
+++ b/recipes-apps/ovmenu-ng-skripts/files/download-xcsoar.sh
@@ -2,18 +2,19 @@
 
 USB_PATH="/usb/usbstick/openvario/download/xcsoar"
 DOWNLOAD_PATH="/home/root/.xcsoar"
+SAVE_FILE="${USB_PATH}/xcs-`date +'%Y-%m-%d'`.tgz"
 if [ ! -d "$USB_PATH" ]; then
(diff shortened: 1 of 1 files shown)


bomilkar:
Here is an example of the issue, just to be more explicit:
if "find $USB_PATH -type f" yields this:
`/usbstick/openvario/upload/xcsoar/Moosburger ALK 2016 07.cup`
then the this
`for uploadfile in $(find $USB_PATH -type f);`
will result in 4 loop transitions:
/usbstick/openvario/upload/xcsoar/Moosburger
ALK
2016
07.cup

This pattern is common to other upload and download scripts. And I don't know how to solve
this keeping the for loop.

recipes-apps/ovmenu-ng-skripts/files/download-xcsoar.sh
```diff
@@ -2,18 +2,19 @@
...
 	mkdir "$USB_PATH"
 fi
+
 if [ -z "$(ls $DOWNLOAD_PATH/* 2>/dev/null)" ]; then
         echo "No files found !!!"
 else
-        for downloadfile in $(find $DOWNLOAD_PATH -type f); do
-                echo $downloadfile
-                cp $downloadfile $USB_PATH
-        done
```
kedder, line 14: I wonder why aren't we just doing `cp -vr $DOWNLOAD_PATH $USB_PATH` here? What is the point of copying files one by one?

bomilkar: I'm happy as long as we get rid of the for loop which breaks with nasty file names.

The advantage of tar is:
it captures a complete snapshot and allows you to retrieve old version of files,
since it's compressed it creates less I/O which seem to be the bottleneck.

Maybe "cp -vr" is a workaround for the other upload and download scripts.

bomilkar: BTW: is automount working? Since which image version?

kedder: I agree. The only disadvantage of tar I see is that the format might not be super-accessible for windows users. That said, I did not use windows for a long long time (maybe 10 years), so my knowledge might be very outdated. Do they support tar.gz on windows natively these days?

kedder: Automount should be working, but it was fixed after 20067, see #15.

bomilkar: 7-zip opens tar.gz files on Windoze. And maybe other tools do, too. This needs to be documented.

bomilkar: Can we drive this to a consensus, please?
Do we use "tar czf" even if the user has to use some other tool to open the tar-file on Win?
Or we use "cp -r" ?

Does anybody have experience with system time on the normal OV machine w/o RTC?
Is that system time really random? Or does it at least advance even if it's not correct?
I'm asking because "cp -ru" could save a lot of copy time, but this is not an option if the system time bounces around.

kedder: I guess I'd prefer `cp -r`, just to keep things as they were, but won't be too upset with tar.gz.

For the RTC: I checked my device yesterday. I don't have battery indeed and clock isn't "ticking" when device is off. However on boot the time is set to last seen value, so we might be able to make an assumption that time is monotonously increasing, even if it is don't match with UTC. I'm not 100% sure about this though, as I didn't check what happens when you do hard reboots. It might be the linux that saves and restores time on reboots.

PR #61, opened 2020-04-29
recipes-core/images/openvario-base-image.bb: add git

Scumi:
Apart from various cloud services, git can be used to manage the xcsoar data folder between different machines. This PR integrates git into the base image.


recipes-core/images/openvario-base-image.bb
```diff
@@ -42,6 +42,7 @@ IMAGE_INSTALL = " \
     nano \
     openssh-sftp-server \
     pulseaudio-server \
+    git \
```
kedder, line 45: I'd recommend to add it to `DEPENDS` variable above instead adding it to image itself. This way it is installable with `opkg` for those who need it, but does not add to image size.

Scumi: Agreed, done.

PR #73, opened 2020-05-23
Tslib support for Touchscreen

linuxianer99:
To enable touch calibration and rotation tslib is used.
It is transparent to the application by using ts_uinput which creates a virtual event device.
The real device has to be disabled for libinput.


linuxianer99:
@kedder : May you have a short look an this for sign-off?

PR #88, opened 2020-06-17
Fix the XCSoar freezing problem

### LINKED ISSUE #71: OpenVario freezes in flight
In 2 of 3 flights with the new image (version 20130 for 7" Texim) I experienced a system freeze which I had never seen with the old image in 300+ hours of flight time. I tried 2 different SD-Cards (with the same image): no difference.

My configuration:
SteFly system with adapterboard, without sensorboard
FLARM on ttyS1, e-vario on ttyS2 ($POV,E,0.58,P,944.33,Q,570.83*1E sentences 2 per second)
sensord and variod disabled with systemctl disable
SteFly stick control, no touch screen

The symptoms:
Acouple of hours into the flight I see that cursor movements (stick control in mouse mode) are getting sluggish (slow reaction). Especially on busy maps, like zoom out to 50km or more.
When moving the mouse on a zoomed-out map the system freezes, meaning no more (visible) map updates, many InfoBoxes still update, but not all. Barometric altitude freezes, for instance.
I leave it alone for ~10 minutes, keeps frozen. Power cycle is the only way to get it going again.
When I end XCSoar after the flight I see a lot of stuff on the screen for a split second before the OpenVario menu takes over the screen again. (I'll change start_xcsoar to capture this "stuff", hoping it reviles something.)
One other thing, which may be unrelated: I find it impossible to "point&click" on the map to get to the details of an airport. That was never a problem with the old image. To some extend it might be related to the "sluggishness" of the cursor movement, but even when it's spot on it doesn't want to pick the airport it is pointing to. I wonder if the "internal pointer on the map" points to the same place as on the screen.
I have the CPU-Load InfoBox open. It seems to update even if the map is frozen. CPU load is reported between 20% and 30% even in freeze. (But the InfoBox may only report user CPU, not system.)

@linuxianer99 , @kedder I think this requires attention. Whoever works/worked on the OS should be alerted. In my humble opinion the system is not usable for the "normal" pilot unless this is fixed.
I have logs from XCSoar both IGC and NMEA which I can make available if that helps. As soon as I have changed ovmenu-ng.sh to capture stderr and stdout from xcsoar I'll supply these, too. Next flight tomorrw :-) :-)

kedder:
Work around the XCSoar random freezing problem, as discussed on [lima/linux#33](https://gitlab.freedesktop.org/lima/linux/-/issues/33). I'm going to submit the patch to to mainline kernel.

As a bonus, remove old unused kernel patches.

FIxes #71


kedder:
There was some positive feedback on XCSoar forum: https://forum.xcsoar.org/viewtopic.php?f=30&t=3669&start=60#p21206

kedder:
@linuxianer99, shall we merge this?

linuxianer99:
@kedder : Yes of course :)

PR #96, opened 2020-07-12
XCSoar patch to set profile selection screen rotation

kedder:
Currently XCSoar will always render the profile selection screen in
landscape orientation, because the actual orientation setting is
specified in profile itself.

These patches are submitted upstream as
https://github.com/XCSoar/XCSoar/pull/453. This commit should be reverted
as soon as it accepted upstream.


kedder:
Closing this because patches were accepted upstream.

PR #116, opened 2021-01-31
including FreeVario driver

Blaubart:


linuxianer99 (approved):
Looks ok for me !

recipes-apps/xcsoar/xcsoar-testing_git.bb
```diff
@@ -52,6 +52,7 @@ SRC_URI = " \
 	file://0001_no_version_lua.patch \
 	file://0001-avoid-tail-cut.patch \
 	file://0007-Disable-touch-screen-auto-detection.patch \
+	file://0001-FreeVario \
```
kedder, line 55: I wonder if it actually builds. There is no file named `0001-FreeVario`. There is one named `0001-FreeVario.patch`. Did you try to build it?

Blaubart: Sorry, wrong file name! I correct it now!

kedder (approved):

PR #118, opened 2021-02-02
Including the new adapter board DS2 and remove machine openvario-7-AM070

Blaubart:
The new adapter board DS2 is a useful further development of the original adapter board. The creation of new machines is necessary because the new adapter board is able to use all four of the ttyS interfaces. This makes it necessary to compile the image a little differently.  In addition, they are new and sensibly arranged. All other changes at the adapter board do not require any software adjustments.

The Ampire display is not compatibel with the original adapter board. That's why I removed it again.


kedder:
Hm, I don't like the combinatoric explosion caused by the introduction of this DS2 (what DS2 means anyway?). This needs to be solved in some different way, not by multiplying machines and duplicating scripts.

The removal of the machine should at least be a separate PR. I'd like it to be discussed with @glidero, who introduced it originally in #113 several days ago.

Blaubart:
> Hm, I don't like the combinatoric explosion caused by the introduction of this DS2 (what DS2 means anyway?). This needs to be solved in some different way, not by multiplying machines and duplicating scripts.
> 
> The removal of the machine should at least be a separate PR. I'd like it to be discussed with @glidero, who introduced it originally in #113 several days ago.

I'm in contact with glidero. I informed him about removing the machine openvario-7-AM070, because it is senseless. The old adapter board dosen't support a 4th LVDS channel! Thats why it is necessary to include the new hardware, that glidero has developed. This adapter board is able to support a 4 LVDS channel display!

glidero:
Hi Andrey,
as Dirk already said, and I agreed, Ampire can not be used on original adapterboard without additional modifications (connecting 4th LVDS channel from cubiboard's U14 connector).
Now what DS2 means. "DS" are my initials. Pretty pathetic, but that's the first thing that crossed my mind. "2" is revision number. Nr. 1 was introduced 2 years ago here - [https://forum.xcsoar.org/viewtopic.php?f=30&t=3371](url) , even it hadn't name DS1. 
Back in DS1 was designed, it was pretty easy to modify image file (17119). All configurations were stored in script.bin/fex file located on FAT sector of microSD. It could be done in notepad without any linux or compilation knowledge. 
DS2 revision was created 1 year ago and 17119 was still the only option for OpenVario. 
Now,  after release of 20193 image, modifications became more complicated because of compilation necessity. 
To keep this new DS2 adapterboard design (and DS1 as well) alive and image updates user friendly, we would like to join the existing philosophy by adding new machines. I also think both designs, official, and DS2, can coexist on one branch. If not, I your advice will be apreciated 
P.S. Of course, I don't insist on this DS2 name. It was more less working name to make obvious the design differences compared to official adapterboard.

kedder:
@Blaubart, @glidero thanks for explanations guys, it makes much more sense now! I have nothing against the name, I simply wasn't familiar with it.

> Back in DS1 was designed, it was pretty easy to modify image file (17119). All configurations were stored in script.bin/fex file located on FAT sector of microSD. It could be done in notepad without any linux or compilation knowledge.
DS2 revision was created 1 year ago and 17119 was still the only option for OpenVario.
Now, after release of 20193 image, modifications became more complicated because of compilation necessity.

It is very similar now as well. Stuff that was stored in `.fex` file now stored in `.dtb` (device tree) file. You can edit it and store on the FAT partition without rebuilding everything. The (significant) difference is that this file is binary, so you would neet some tools to convert it to text and back (dtc - device tree compiler).

Now, I myself am fine with introducing other targets. The changes make sense except the duplication of menu script for each architecture. OTOH it is not your fault. For some reason it was like that before.

Finally, currently the PR is not mergeable - there are conflicts with the other changes that were recently merged. You would need to rebase your branch on top of current `warrior` branch, resolve conflicts and force push it.

Blaubart:
Sorry, Andrey, I think I did it wrong. Iv tried to resolve the conflict and I deleted my fork, made a new one and made my changes again. But this fork is not combined with my pull request. It seems the only option I have is to make a new PR, but thats not good option.

kedder:
@Blaubart no problem. There is a way to keep the same PR by pushing to the branch that is already under review, but I guess it is hard to grasp without some experience with git. Let's continue on #119.

PR #122, opened 2021-02-22
Disable SPI, enable ttys4 on UART6

DanD222:
In the current state, the first OpenVario RJ45 socket is currently used as ttyS0, a Linux serial console for debugging, and it is not recommended to connect an IGC-compatible device to it. This means that only 3 IGC-compatible devices can connect to OpenVario.

This modification assigns Cubieboard pins which are currently designated as SPI pins to work as UART pins, so that a fourth IGC-compatible device can be connected to OpenVario using the hardware modification described below. 

As-is:
At present, pins 46 and 48 of Cubieboard connector Cubie1 (U14 in the Cubieboard datasheet) are used for SPI0-MOSI and SPI0-MISO. These are taken to the Sensorboard via the UEXT connector, but are then unused, as **no devices currently interface to OpenVario through SPI**.

To be:
Pins 46 and 48 are assigned to UART6, a fourth IGC-comaptible device can be connected to OpenVario, provided that the following hardware modifications are performed (referenced to the original OpenVario adapterboard):
- Remove the two wires running from the Rx and Tx pins on the Cubieboard to the UART0 connector on the adapterboard _(this breaks the ttyS0 debug feed to the first RJ45 socket)_ 
- Link pin SPI0-MOSI on adapterboard UEXT connector with pin UART1-TX on adapterboard UART0 connector _(this connects UART6 TX to the first RJ45 socket)_
- Link pin SPI0-MISO on adapterboard UEXT connector with pin UART1-RX on adapterboard UART0 connector _(this connects UART6 TX to the first RJ45 socket)_

The IGC device will then be accessible in XCsoar under ttyS4

### DIFF
--- a/recipes-kernel/linux/linux-mainline/openvario-7-AM070-DS2.dts
+++ b/recipes-kernel/linux/linux-mainline/openvario-7-AM070-DS2.dts
@@ -27,10 +27,3 @@
 &uart5 {
     pinctrl-0 = <&uart5_pi_pins>;
 };
-
-// UART6 on Schematic
-&uart6 {
-    pinctrl-names = "default";
-    pinctrl-0 = <&uart6_pi_pins>;
-    status = "okay";
-};
--- a/recipes-kernel/linux/linux-mainline/openvario-7-CH070-DS2.dts
+++ b/recipes-kernel/linux/linux-mainline/openvario-7-CH070-DS2.dts
@@ -27,10 +27,3 @@
 &uart5 {
     pinctrl-0 = <&uart5_pi_pins>;
 };
-
-// UART6 on Schematic
-&uart6 {
-    pinctrl-names = "default";
-    pinctrl-0 = <&uart6_pi_pins>;
-    status = "okay";
-};
--- a/recipes-kernel/linux/linux-mainline/openvario-common.dts
+++ b/recipes-kernel/linux/linux-mainline/openvario-common.dts
@@ -101,6 +101,13 @@
 	status = "okay";
 };
 
+// UART6 on Schematic - enabling this removes SPI functionality in order to enable 4th IGC port usage
+&uart6 {
+    pinctrl-names = "default";
+    pinctrl-0 = <&uart6_pi_pins>;
+    status = "okay";
+};
+
 &rtp {
 	allwinner,ts-attached;
 	allwinner,tp-sensitive-adjust = <0>;


recipes-kernel/linux/linux-mainline/openvario-43-rgb.dts
```diff
@@ -26,3 +26,10 @@
 	pins = "PI10", "PI11";
 	function = "uart5";
 };
+
+// UART6 on Schematic - enabling this removes SPI functionality in order to enable 4th IGC port usage
+&uart6 {
+    pinctrl-names = "default";
+    pinctrl-0 = <&uart6_pi_pins>;
+    status = "okay";
+};
```
kedder, line 35: Do you really need to copy-paste this in all `dts` file? Can we put it to `openvario-common.dts` instead?

kedder (changes requested):
This looks good to me. But we need to figure out how to preserve the instructions about the board modifications after this PR is gone.

And get rid of code duplications in dts.

recipes-kernel/linux/linux-mainline/openvario-43-rgb.dts
DanD222: OK, 

- Moved the code to openvario-common.dts
- The description/instructions are in the commit message

kedder (commented):
Hm, btw, does this hardware modification actually disables the debug port (/dev/ttyS0)? You cannot use the `/dev/ttyS0` as linux console after this, right?

DanD222:
The debug port doesn’t get disabled, it’s still active on the Rx / Tx Cubieboard pins. It just doesn’t get routed to the outside of the case through the IGC connector if you remove the wire from the Cubieboard to the adapterboard.

kedder (approved):
I see. There might be a simpler way to use that extra UART port for the peripherals - simply by not directing the linux console to `/dev/tty0` (I think that's a boot parameter to a kernel). Anyway, this looks good to me.

linuxianer99 (approved):

kedder:
@DanD222, were you able to build the image with this change? For whatever reason the code was indented with "no-break-space" characters (U+00A0), which is a syntax error. Fixing this in #126.

DanD222:
Sorry - I didn’t actually, I didn’t expect something like this to come up.. lesson learned :(

PR #125, opened 2021-03-05
Removed patch file hunks

Blaubart:


linuxianer99:
@Blaubart : I assume XCSoar is still building fine with this changes ??

Blaubart:
Yes, lines have been shifted in the files to be patched. I only adjusted that in the patches. This is not really necessary because the patch still works. But now without the annoying warnings!

kedder (approved):
Looks good to me!

PR #135, opened 2021-05-23
Fix download-all.sh

sfalmo:
Tested on a Stefly 5.7" OpenVario.

When you select the Download option in the ov-menu, no files are transferred to the USB stick.
This is for two reasons:
 - A typo in a shell variable
 - A failing mkdir command

This PR fixes both, and the contents of the .xcsoar directory are transferred to the stick.
Note that this takes quite some time as also the map files are copied.


kedder:
Yeah, that's embarassing. Thanks for the fix.

kedder (approved):

PR #151, opened 2021-12-13
Readme adjustments

August2111:
Yesterday a new version of XCSoar is published - so OpenVario should switch to this!

Unfortunately with patch FreeVario neither the normal package 'openvario-image' nor the testing package are failureless compilable, so we should temporarely switch off - at least for the normal standard package!

Besides that a better name for the sdcard is proposed ('openvario-7' give the user not so much information) and the sensord-testing.bb is better aligned to sensord.bb (better comparison)


August2111:
Wrong PR ;-(

PR #155, opened 2021-12-23
Update sensord

August2111:
Update SensorD fron version 0.3.4 to 0.3.5

Solve the sensord issue  37  'voltage reported by openvario is too high' in a first shot... (better solution in upcoming version)

Additionally I removed all unused variod and sensord recipes: All this stuff is saved in git!!


August2111:
@linuxianer99: Sorry - I made the same change as you  did - but additionally I removed all unused old recipes - I only saw your PR #154 today ;-(

August2111:
Ok, I will close this PR because the main intention is doubled to #154.
The cleaning up I will propose in a next PR...

PR #159, opened 2022-01-04
Improve SDCard image naming

linuxianer99:


linuxianer99:
@August2111 : I grep your changes regarding SDCard Stuff and created a PR how i would do it ...
Can you please have a look ...

August2111 (commented):
Ok, this was the way that kedder proposed

PR #164, opened 2022-01-04
sdcard_image-openvario.bbclass - updating comments

August2111:
This was still missing in PR #159 

The more important thing is the typing error in line 144


linuxianer99:
Just a hint ... There is no need to mention the filename in the commit message. It is obvious, if you look on the commit ...

August2111:
> Just a hint ... There is no need to mention the filename in the commit message. It is obvious, if you look on the commit ...

I know - but so I see the filename in a commit table, and so I know faster where a have to search ...

Issue #156, opened 2022-01-01
Home folder in Openvario


lordfolken:
Hello, 

The current image seems to use /root/.xcsoar for xcsoar's data dir. 
The backup scripts etc point to /home/root/.
https://github.com/Openvario/meta-openvario/blob/cb717925100ca2927ae626084b26710355f67ba7/recipes-apps/ovmenu-ng-skripts/files/download-all.sh#L4

Could we use the $HOME variable?

August2111:
For this data directory I have an other proposal, but this have to be discussed here: Of course you should use the $HOME variable itself - my proposal is to use for this the free area of sdcard as 3rd partition (/dev/sda3/XCSoarData) with FAT32 as standard folder, this is big enough for all files - and you can access to it if you have the sdcard outside of the OV at a Linux or Windows PC!

But of course the DOWNLOAD_PATH and the xcsoar's data dir have to be aligned to each other!

lordfolken:
I would not. Having yet another way to access the xcsoar data folder that works only half the time isn't helping. 

There has to be a clear designed official way. 

Besides:
* fat32 is corruptable, not journaling, not easly fscked etc. 
* fumbeling with micro-sdcards in the cockpit is just unpractical. (hidden slots, things can fall into cracks, the ejection mechasims break, easly lost, adapters, different standards...) ( i know the big ones do it, but at least in my club we are mounting usb ports everywhere now..)
* USB Sticks are much more practical for this

However I would:
 * seperate the userhome from the os:
   * replace the os without wiping the user data
   * should run as a dedicated user, not root
   * filing the data partition won't affect the OS

kedder:
I'm with @lordfolken on this. I'd wish we could utilize a separate data partition on sd card that would use the whole size of an sd card. In a similar fashion ubuntu live-disk works. However I'm not aware of any avarege-user-accessible tools that could create such a filesystem on a flash card.

lordfolken:
RPI do it simillarly. They create a small second paritition, and then resize it at boot time to the whole device. Such a thing could be achieved with sfdisk.

August2111:
Hi kedder, I had done some experimental work on this topic to create a separate partitition on the 4, 8, 16, 32 GB sized sd card: a lot of room for many data. But why lordfolken decline the fat32? From my point of view I don't need the journaling and fsck on my user data, most of the sd cards and usb sticks are formatted with fat32 - and it is easy readable and editable as well on Linux as Windows.
[...]

PR #166, opened 2022-01-04
Reducing the image size

August2111:
Reduce the image size to 1GB for the possible usage of 1GB sd card!

- Maybe this should be aligned later to a common sd card size


kedder:
I would not do this, because when OV goes out-of-space, XCSoar crashes. And I had such an event midflight during WGC2021 :) Takes an extraordinary amount of hacking skills to diagnose the situation and work around it using only the arrow buttons on the stick :)

Out-of-space can happen pretty easily if you have NMEA logging on.

August2111:
> Out-of-space can happen pretty easily if you have NMEA logging on.
Of course there is a good point - I know! But for this reason I want to discuss about the location, access rights, size and so on at issue #156. With big maps - without cleanup between you can also running in this out of space very fast!
At least a warning should be happens, if the memory has only a free space of some megabytes...
So I will postbone this PR after this discussion!

MaxKellermann:
This change is bad; instead of hard-coding an image size, the filesystem should be expanded to the full size of the SD card on first boot. This can be done easily with [systemd-repart](https://www.freedesktop.org/software/systemd/man/systemd-repart.html).
Also, there should be separate partitions for the operating system and data, to allow updating the firmware without erasing all data.

DanD222:
Pardon me for being obtuse, but what is the benefit of reducing the image size to fit on a 1GB SD card in an age when high capacity SD cards are cheap and plentiful?

kedder:
@DanD222 as I understand the main benefit is quicker turn-around when a developer tries a new build. Flashing 1GB is 4 times faster than flashing 4GB and you have to do it a lot of times when debugging the image. Our current set up limits the available free space to 1GB though, so it is not great. @MaxKellermann basically suggests the same thing we have agreed on in #156, which is a win-win - making image small and quick to flash, and also using all available flash space.

DanD222:
@kedder OK, that sounds good, thanks for the explanation!

MaxKellermann:
The image could be just 50 MB (with a compressed read-only system partition). The data partition spanning the rest of the SD card would then be initialized on the first boot. You can flash an OpenVario within a couple of seconds that way...

August2111:
Ok, all this stuff was the main idea behind this commit/pr: reducing the os partition and create a big data partition, so we discussed in #156 this idea - and at the end we will bring this ideas together in a new PR - so I will close this one to make the way free for the next steps...

PR #167, opened 2022-01-04
Linux kernel 5.10.2

August2111:
Switch to Linux kernel 5.10.2 (from outdated 5.7)
- use the original layer from yokto/poky, openembedded and sunxi as it is - and make small modifications needed by OV only

- adding additional drivers will coming in a next step/PR


kedder (approved):
Great!

PR #168, opened 2022-01-04
fix xcsoar builds

Scumi:
XCSoar does not build anymore after the last set of merges (missing patch for testing, wrong format for production build). Those changes fix this.


August2111:
Thanks to @Scumi: I'm really shure that originally a correct number of spaces was in - but it disappears because some splitting and copying actions.. ;-( 
But now it is fine - very good (and very fast)!

Scumi:
I was excited to re-build the image with your changes in and it didn't work! :smiley: 
Luckily I found the issue pretty fast, many thanks for your extensive work on this lately!

Issue #89, opened 2020-06-18
Problem with rotary encoder


Blaubart:
I installed the last self compiled image from yesterday where touchscreen is working. If I choose NMEA Device at menu, I can select device with up and down (rotary encoder at the right hand side), but I can't choose edit, monitor, reconnect etc. (rotary encoder at the left hand side). At other menu options right and left with this encoder works. With the old image everything works fine.

kedder:
This might be an XCSoar bug. What version of XCSoar are you running? Blue one (6.x) or Red one (7.x-preview)?

Blaubart:
7.0_preview15

Blaubart:
of course, I will try!

Blaubart:
I tried an image from today with XCSoat 6.8.x. It is the same

kedder:
I just started the old image out of curiosity,  and it seems like it behaving exactly the same way as on new image.

I don't have rotary encoder, but I'm assuming it is just a usb keyboard device, producing arrow keypresses when you rotate knobs. If it is so, on old image (17119) it is possible to select device with up/down keys, but left/right button just make cursor jump to first and last device (you still can't choose edit, monitor, reconnect etc.). Maybe your rotary encoder produces some other keypresses?

Blaubart:
rotary encoder just producing arrows, yes. Up and Down works perfectly. left/right button just make cursor jump to first and last device. It's exactly the same at my and your device.

kedder:
But that is the same on old image too.  I think this is XCSoar issue and it ptobably never worked as you are expecting it to work. I think this should be filed as a bug on XCSoar project. There is not much openvario can do about it.

Blaubart:
I tried 4 different images:
- 17119 but with a self compiled XCSoar 7 (everything works)
- a selbst compiled a couple weeks old with XCSoar 7.0_preview15. Sorry, don't know exactly how old it is (everything works)
- image from yesterday with XCSoar 7.0_preview15 (dosen't work)
- image from today with XCSoar 6.8.14 (dosen't work)

kedder:
Does it work for you with normal usb keyboard (not rotary encoder)?

Blaubart:
with a keyboard it is exactly the same.

kedder:
Weird, I can't make it work on old image with my keyboard

kedder:
[...]

PR #170, opened 2022-01-08
fix for Rotary encoder

Blaubart:
The rotary encoder is now working again to control the respective items in the menu. Solves this problem:

https://github.com/Openvario/meta-openvario/issues/89

### DIFF
--- a/recipes-apps/xcsoar/files/0001-HasNoPointer.patch
+++ b/recipes-apps/xcsoar/files/0001-HasNoPointer.patch
@@ -1,0 +1,23 @@
+--- a/src/Form/ButtonPanel.cpp
++++ b/src/Form/ButtonPanel.cpp
+@@ -353,7 +353,7 @@ ButtonPanel::KeyPress(unsigned key_code)
+     }
+   }
+ 
+-  if (selected_index >= 0 && !HasPointer()) {
++  if (selected_index >= 0) {
+     if (key_code == KEY_LEFT) {
+       SelectPrevious();
+       return true;
+       
+--- a/src/Form/Button.cpp
++++ b/src/Form/Button.cpp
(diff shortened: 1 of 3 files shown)


linuxianer99 (approved):

linuxianer99:
@MaxKellermann:  Does this patch make sense ??

MaxKellermann:
Somewhat related bug report: https://github.com/XCSoar/XCSoar/issues/625

This change is a dirty kludge, which undoes a side effect of https://github.com/Openvario/meta-openvario/blob/warrior/recipes-apps/xcsoar/files/0007-Disable-touch-screen-auto-detection.patch ... which already was a dirty kludge. So it's one dirty kludge stacked on top of another dirty kludge.
Such things don't make me happy. I must admit that the XCSoar input code is *very* far from perfect, but this isn't a proper solution. This may be short-term ok to merge for OpenVario only, but OpenVario should keep its set of patches as small as possible, and we better search for a general solution for all targets.

linuxianer99:
OK ... think i got the point.
As there is already an open issue in the XCSoar project (XCSoar/XCSoar#625), we can merge this for the moment.

@Blaubart can you please work on XCSoar/XCSoar#625 to find a better solution ??

@kedder Think we can merge ...

kedder:
@MaxKellermann, I agree, these patches are not great. The reason they exist is nobody could come up with a generic solution to these problems that would fit into mainline XCSoar. I can see how changes in this behavior can affect users on devices I might not even know about. 

OV kinda have a pointing device, but it is not primary way of input and it is there for "backup" (when navigating with arrows is not possible). The joystick emulates a mouse, but actually expected to behave more like a touch input (less precise), because precise clicks are very hard to do in flight. 

This breaks assumptions made in XCSoar, but I suppose these assumption are not made for no reason and some users depend on them.

PR #172, opened 2022-01-10
Remove some bloat (unnecessary dependencies and build options)

MaxKellermann:
This reduces the firmware size from 143 MB to 132 MB. A lot more is possible; for example, PulseAudio is responsible for a very long rat tail of dependencies.


linuxianer99 (approved):
Looks resonable !

Thanks, Max

kedder (approved):
This looks good to me (as long as it works). I cannot spot anything that would break any supported features.

@MaxKellermann, we use pulseaudio for mixing sounds emitted by xcsoar and variod (built-in beeping variometer). I'm not competent enough in linux audio subsystem to tell if it can be achieved with simpler set up, but pulseaudio worked pretty well for us.

MaxKellermann:
XCSoar has a beeping variometer as well, so this is kind of redundant. But if you really want to share the audio device between two processes, you can use ALSA dmix instead. PulseAudio is big fat overkill for this.

PR #173, opened 2022-01-12
Various XCSoar build improvements

MaxKellermann:


recipes-apps/ovmenu-ng/files/openvario-43-rgb/ovmenu-ng.sh
```diff
@@ -342,9 +342,9 @@ function upload_files(){
 function start_xcsoar() {
 	/usr/bin/xcsoar_config.sh
 	if [ -z $XCSOAR_LANG ]; then
-		/opt/XCSoar/bin/xcsoar -fly -480x272
+		/usr/bin/xcsoar -fly -480x272
```
kedder, line 345: Note to self: this will also requires an update to openvario-shell.

kedder (approved):
All looks good to me. Was easy to review, thanks for well separated commits.

linuxianer99 (approved):

Issue #24, opened 2020-03-12
Include nano editor


mihu-ov:
I know that real Linux pros use vim but for the average user trying to modify config files nano would be much easier to use. We had it in the old image but it is not included in 20067.

PR #174, opened 2022-01-13
More debloating

MaxKellermann:
Reduces the rootfs size from 132 MB to 124 MB.

### DIFF
--- a/recipes-core/images/openvario-base-image.bb
+++ b/recipes-core/images/openvario-base-image.bb
@@ -2,8 +2,7 @@ SUMMARY = "A small image just capable of allowing a device to boot."
 
 IMAGE_FEATURES += "splash ssh-server-dropbear package-management"
 DEPENDS += "linux-firmware \
-    git \
-"
+    "
 
 # Include common WIFI firmware packages into the image. All linux-firmware
 # packages take almost 500MB of space uncompressed, so we don't want to ship
--- a/recipes-core/images/openvario-base-image.bb
+++ b/recipes-core/images/openvario-base-image.bb
@@ -41,7 +41,6 @@ IMAGE_INSTALL = " \
     autofs-config \
     nano \
     openssh-sftp-server \
-    pulseaudio-server \
     tslib \
     tslib-tests \
     tslib-uinput \
--- a/conf/distro/ovlinux.conf
+++ b/conf/distro/ovlinux.conf
@@ -52,7 +52,8 @@ DISTRO_EXTRA_RRECOMMENDS += " ${POKY_DEFAULT_EXTRA_RRECOMMENDS}"
 
 DISTRO_FEATURES ?= "${DISTRO_FEATURES_DEFAULT} ${DISTRO_FEATURES_LIBC} ${POKY_DEFAULT_DISTRO_FEATURES}"
 DISTRO_FEATURES_append = " systemd"
-DISTRO_FEATURES:remove = "acl argp pcmcia usbgadget xattr nfs zeroconf pci 3g nfc x11"
+DISTRO_FEATURES:remove = "acl argp pcmcia usbgadget xattr nfs zeroconf pci 3g nfc x11 gobject-introspection-data"
+
 VIRTUAL-RUNTIME_init_manager = "systemd"
 
 # Make /var/log dir persistent on the filesystem. By default logs are located
--- a/conf/distro/ovlinux.conf
+++ b/conf/distro/ovlinux.conf
@@ -38,6 +38,10 @@ PACKAGECONFIG:remove:pn-openssh = "rng-tools"
 PACKAGECONFIG:remove:pn-glib-2.0 = "libmount"
 SHAREDMIMEDEP:pn-glib-2.0 = ""
 
+PACKAGECONFIG:remove:pn-libsndfile1 = "alsa"
+DEPENDS:remove:pn-libsndfile1 = "flac libogg libvorbis"
+EXTRA_OECONF:append:pn-libsndfile1 = "--disable-external-libs"
+
 PACKAGECONFIG:append:pn-wget = " gnutls"
 PACKAGECONFIG:remove:pn-wget = "openssl pcre"
 
--- a/conf/distro/ovlinux.conf
+++ b/conf/distro/ovlinux.conf
@@ -48,6 +48,9 @@ PACKAGECONFIG:remove:pn-wget = "openssl pcre"
 
 PACKAGECONFIG:remove:pn-alsa-plugins = "samplerate speexdsp"
 
+# building nano without libmagic omits the large "magic.mgc" file
+DEPENDS:remove:pn-nano = "file"
+
 PREFERRED_VERSION_linux-yocto ?= "4.18%"
 
 SDK_NAME = "${DISTRO}-${TCLIBC}-${SDK_ARCH}-${IMAGE_BASENAME}-${TUNE_PKGARCH}"
--- a/conf/distro/ovlinux.conf
+++ b/conf/distro/ovlinux.conf
@@ -35,6 +35,7 @@ PACKAGECONFIG:remove:pn-connman = "wispr"
 PACKAGECONFIG:remove:pn-curl = "libidn proxy verbose"
 PACKAGECONFIG:remove:pn-bluez5 = "obex-profiles readline a2dp-profiles avrcp-profiles network-profiles hid-profiles hog-profiles deprecated"
 PACKAGECONFIG:remove:pn-openssh = "rng-tools"
+PACKAGECONFIG:remove:pn-gnutls = "libidn"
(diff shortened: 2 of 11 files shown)


recipes-core/images/openvario-base-image.bb
```diff
@@ -39,7 +39,6 @@ IMAGE_INSTALL = " \
     wget \    
     autofs \
     autofs-config \
-    nano \
```
kedder, line 42: This is controversial because it reverts #24.

recipes-core/images/openvario-base-image.bb
```diff
@@ -2,8 +2,7 @@ SUMMARY = "A small image just capable of allowing a device to boot."
 
 IMAGE_FEATURES += "splash ssh-server-dropbear package-management"
 DEPENDS += "linux-firmware \
-    git \
-"
+    "
```
kedder, line 5: I think the idea was to allow users to manage their .xcsoar dir with git (if they want to).  That said, I'm not sure this has to be in the base image. I'm fine with removing it. 

@Scumi, if you still need git on openvario, I can make it installable through my https://github.com/kedder/openvario-ppa.

recipes-core/images/openvario-base-image.bb
MaxKellermann: The general question is: what should the OpenVario image that everybody uses really contain? I think it should be as light as possible, and we should even get rid of bash completely (because busybox should be enough).
You could provide an extended "developer" or "power user" image with more tools, but don't bloat the basic user image with stuff 99% don't need.
If nano remains in the basic image, we could/should eliminate the "magic.mgc" dependency.

recipes-core/images/openvario-base-image.bb
MaxKellermann: But ... who really does that? Even I wouldn't do that - and even if, I would rsync to my main computer and have the git repository only there.
So if we're talking about useful tools for power users, rsync would be one, but not git.

Scumi: I'm fine with this change. It wasn't practical.

recipes-core/images/openvario-base-image.bb
kedder: Yeah, this is a topic for a broad discussion. I'll let @mihu-ov and @linuxianer99 correct me if I'm wrong, but my impression is that conceptually OV is not just a typical consumer device, but a tinkerer-friendly one. So it includes some quality-of-life tools that tinkerers would benefit from.

We can make them installable separately with opkg and some internet-hosted ppa (like the [one I'm hosting](https://github.com/kedder/openvario-ppa)), but to use it you need to have internet on your OV box. And to have it, currently you need to buy a wifi dongle separately, have an usb keyboard attached and configure wifi through command line.

We can have a separate images for "consumer" and "tinkerer", consumer being a "bare minimum". This is also not straightforward. We already providing a separate image for each "display size" flavor of OV device. That number is multiplied by "stable/testing" flavors. "consumer/tinkerer" would multiply that number further by 2. It would be 32 or so images to build for each release. If we go this route,  we would have to come up with something to prevent this combinatoric explosion.

Furthermore, I'm making a thing called "[openvario-shell](https://github.com/kedder/openvario-shell)", which makes OV UI a bit friendlier and makes it simpler to write additional software for the device. It also includes tools to configure the wifi, so you don't have to go to command line to do that. However, it is written in python (specifically to make it simpler to develop). And if it will ever be included in stock OV image, it will pull all the python runtime with it, which might nullify your efforts cutting the image size down :) Python is a pretty heavy dependency FWICT.

MaxKellermann: Oh, those "display size" flavors ... I saw the horrific shell code duplication in this git repo, a 400 line shell script monster of which there are 7 copies, the delta being just one line with the screen resolution. How are you able to manage this unnecessary complexity?

Also I don't understand why different images are needed for different display sizes. Another unnecessary complexity disaster! If that cannot be auto-detected, it could be a system configuration; start with a fallback resolution that works everywhere, and allow users to switch on the first boot.

And please don't include Python in the image. It's really a huge dependency which will probably double the image size.

Why not write a C++ program similar to KoboMenu (which does the same thing on the Kobo - system configuration and XCSoar launcher), and manage it inside the XCSoar git repository? That way, it can use the XCSoar GUI toolkit, all the infrastructure, and we have a consistent user interface for everything.

kedder: > a 400 line shell script monster of which there are 7 copies, the delta being just one line with the screen resolution

Yes, that one. Not sure how/why that happened, but my intention was to eventually replace it with openvario-shell. So I never attempted to tide it up.

> Also I don't understand why different images are needed for different display sizes

The main reason is that display configuration (all the frequencies for the LVDS display driver) need to be available on boot (i.e. `u-boot` need to know them) and they are not really autodetectable from my understanding. Those are constants that are being compiled-in into `u-boot`. There's also a separate device-tree with those constants for the linux kernel. This forces us to ship a separate u-boot package (and effectively the whole separate kernel package, even though the only difference is the device-tree file) for each display size.

> Why not write a C++ program similar to KoboMenu

My main motivation for buying OV in the first place was to be able to program it. I.e. [openvario-compman](https://github.com/kedder/openvario-compman). I chose python to make the software easily hackable on your laptop. . Python allows to do it relatively easily: you can program and debug it on your laptop and then just copy files over to your OV device and run it. Cross-compiling for OV is a completely different story (maybe that's just me not familiar with C++ tools though). "Go" would be another good choice IMO (it produces statically linked native binaries), but it'd be too much effort to switch at this point.

In any case, I'm not pushing python into OV at this point, however I think it would be a good thing to ship with the openvario: ability to script the device is a very powerful one IMO.

MaxKellermann: You could have one image which contains many DTs and many u-boots, and write a menu to switch (and reboot).
The Go language is another bloat desaster. The typical Go program weighs more than the whole OpenVario image ...
If you want scripting: XCSoar comes with a full Lua scripting engine, which weighs 200 kB or so.
About debugging on your laptop: you can debug KoboMenu on your laptop, and the same way, you could debug a C++ OVShell on your laptop.
The KoboMenu executable weighs less than 1 MB, and that includes a copy of the whole of XCSoar's GUI toolkit (statically linked). That's much more than the current set of shell scripts, but the shell script has dependencies, too, which we could eliminate, so it would be a net win.

linuxianer99:
Hey All,
so first of all, thanks for the valuable input and all the good discussions!
I agree totally with @MaxKellermann that there are lot of possible improvements !

Maybe just to explain, why some of the code is so "horrible" ... OV was initially developed by three persons only. So Andreas and Michael did the mechanical part and i did the software part. I was also not that experienced at the start ...
@MaxKellermann your input is very welcome, as it would be great, if we can bring Openvario to a similar code quality as XCSoar is !!

As time goes by, other people also joined the development, but most of them also have no software background so everybody tried his best to contribute. Think that is the main reason ...

From this PR is see, that there are two major topics we have to discuss in general:

1.  Strategy for the image in regard to "production", "Poweruser", "developer" and thee corresponding content of it.
2. How to reduce or refactor some of the software 

If it is ok for you, i will start two discussions in the discussion tab of the project ?

recipes-core/images/openvario-base-image.bb
linuxianer99: For me it would be okay, to remove nano ...

recipes-core/images/openvario-base-image.bb
linuxianer99: Would be also ok to remove git and think about a "development image" or a possibility to install the tools afterwards ...

linuxianer99:
First discussion started #181

recipes-core/images/openvario-base-image.bb
mihu-ov: I understand the goal to shrink the image but I´d rather keep nano in the image to edit config files etc. Not many people on an airfield are comfortable with vim or something the like.

MaxKellermann: I'll try to find a way to make nano lighter, without the magic file.

MaxKellermann: Done and force-pushed. nano itself weighs only 230 kB, but omitting the useless-for-us magic.mgc file saves 7 MB of image size.

mihu-ov: Thank you!

linuxianer99: Thanks @MaxKellermann ! Very good solution !

recipes-core/images/openvario-base-image.bb
linuxianer99: Think we can go without git ...

linuxianer99 (approved):

kedder (approved):

PR #186, opened 2022-01-15
linux-mainline: convert arm/defconfig to a delta .cfg file

MaxKellermann:


linuxianer99 (approved):
Looks good to me!

kedder (approved):
LGTM!

PR #188, opened 2022-01-17
Kernel configuration tweaks

MaxKellermann:


linuxianer99 (approved):

MaxKellermann:
Don't merge yet - I found a problem with the config analysis step (and a Yocto bug crashing Bitbake) - will force-push when it's solved.

MaxKellermann:
Workaround in place (`KMETA`) - I think this can be merged now.

recipes-kernel/linux/linux-mainline/openvario.cfg
```diff
@@ -105,19 +99,13 @@ CONFIG_SND_AUDIO_GRAPH_CARD=m
 # CONFIG_HID_REDRAGON is not set
 # CONFIG_HID_MICROSOFT is not set
 # CONFIG_HID_MONTEREY is not set
-CONFIG_USB_ANNOUNCE_NEW_DEVICES=y
```
kedder, line 158: I think this is a useful thing to keep enabled. Especially when someone complains their device doesn't work. We at least can undersdtand what kind of device is that from the log.

recipes-kernel/linux/linux-mainline/openvario.cfg
```diff
@@ -105,19 +99,13 @@ CONFIG_SND_AUDIO_GRAPH_CARD=m
...
 CONFIG_USB_XHCI_HCD=m
 CONFIG_USB_XHCI_PLATFORM=m
 CONFIG_USB_MUSB_SUNXI=y
-CONFIG_USB_ACM=m
```
kedder, line 162: Are you sure this will not break USB-serial devices that register as `/dev/ttyACM0`?

recipes-kernel/linux/linux-mainline/lima.conf
```diff
@@ -1,3 +0,0 @@
-# lima kernel module configuration
-softdep lima pre: sun4i-drm sun4i-drm-hdmi sun4i-frontend sun4i-tcon sun4i-backend
-options lima job_hang_limit=10
```
kedder, line 3: The patch was deleted because it was incorporated into the mainline linux. This configuration is now understood by the kernel without the special patch. This is still an important piece of configuration. Without it the device hangs occasionally.

recipes-kernel/linux/linux-mainline_%.bbappend
```diff
@@ -17,6 +17,7 @@ SRC_URI_append = " \
         file://usbnet.cfg \
         file://usbserial.cfg \
         file://filesystems.cfg \
+	file://lima.cfg \
```
kedder, line 21: I'd call it `graphics.cfg` - most of the options doesn't seem to be directly related to lima driver.

recipes-kernel/linux/linux-mainline/openvario.cfg
```diff
@@ -1,3 +1,4 @@
+CONFIG_DEFAULT_HOSTNAME="openvario"
```
kedder, line 1: Ah, that's how you do it!

kedder (changes requested):

recipes-kernel/linux/linux-mainline/openvario.cfg
MaxKellermann: The image contains `lsusb` which is much more useful, because prints more information. The kernel log is very brief.

recipes-kernel/linux/linux-mainline/openvario.cfg
MaxKellermann: Which one does? Does it pretend to be a USB modem, instead of a serial port? (I asked about USB modem support on https://github.com/Openvario/meta-openvario/discussions/185)

recipes-kernel/linux/linux-mainline/lima.conf
MaxKellermann: The presence of this file causes a build breakage, because it is not included in any binary package. So either delete it (what this commit does), or add it to the FILES variable of a package.

MaxKellermann: After https://github.com/Openvario/meta-openvario/pull/188/commits/dbfcb6e1286c24435befaddd45bfdf8e609f2735 the module parameter must be moved to the kernel command line.

recipes-kernel/linux/linux-mainline/openvario.cfg
kedder: At least one device I can recall that does that are Arduino boards. Not sure why they do that - they are not modems, but appear as `ttyACM0` when you connect them. I think I saw more of them but can't recall what it was. To me it's a mystery why some USB-serial devices are `ttyUSB0` and some `ttyACM0`, but I think both are pretty useful for OV.

MaxKellermann: OK, dropping this part in the next push.

recipes-kernel/linux/linux-mainline/lima.conf
MaxKellermann: Dropped for now (which keeps the build breakages), and I'll later figure out how to add this to the kernel command line.

recipes-kernel/linux/linux-mainline_%.bbappend
MaxKellermann: Renamed.

recipes-kernel/linux/linux-mainline/openvario.cfg
kedder: I often rely on the kernel logs to determine if device is working / recognized or not. If there is some blurp in the logs when I plug in the device, at least I know it is powered and talks to the host. Dunno, this feature helped my a lot in the past, I'd keep it.

MaxKellermann: Changed so the option is kept.

kedder (approved):

recipes-kernel/linux/linux-mainline/lima.conf
kedder: For the record, the problem is documented #71, the epic debugging effort.

PR #190, opened 2022-01-18
Some systemd goodies (and some recipe cleanup)

MaxKellermann:


linuxianer99:
Looks good to me! 
Thanks for the improvements !!

Just a question:
Did you have a change to test this automount as well as auto login ?? You mentioned you have no CB2 at home ...

linuxianer99 (approved):
Like to merge !

recipes-support/ovmenu-ng-autostart/ovmenu-ng-autostart/ovmenu-ng.service
```diff
@@ -0,0 +1,45 @@
...
+Conflicts=rescue.service
+Before=rescue.service
+
+# On systems without virtual consoles, don't start any getty. Note
+# that serial gettys are covered by serial-getty@.service, not this
+# unit.
+ConditionPathExists=/dev/tty0
+
+[Service]
+ExecStart=/opt/bin/ovmenu-ng.sh
```
kedder, line 28: Something that we need to make sure is working: ovmenu has "exit to shell" function. The way it works now (if I'm not mistaken) is ovmenu simply exits and a user drops into normal bash prompt. When user exits bash, the ovmenu is getting restarted by autostart (and user sees the menu again).

Does this still work with systemd service?

kedder (approved):
Looks good otherwise! Nice cleanup!

recipes-support/ovmenu-ng-autostart/ovmenu-ng-autostart/ovmenu-ng.service
MaxKellermann: No, indeed, that's broken now! I suggest removing that menu item, and instead going to tty2 with Alt-F2, what do you think? This "standard" way of switching ttys was broken previously, but now works perfectly.

kedder: What about just running `bash` instead of exiting the ovmenu? will it work with this setup?

MaxKellermann: That should work, I just thought it was easier for Linux guys to just switch to another VT.

kedder: Alt-F2 works for me, we would just have to answer questions from people who will miss that feature. Otherwise shell prompt is useless without a keyboard and all keyboards have Alt and F2.

linuxianer99: i agree with you @MaxKellermann and @kedder . 
Every keyboard has Alt-F2 :-)

MaxKellermann: So.... remove the menu item or fix it?

MaxKellermann: As a hint ... XCSoar shouldn't be running as root, and neither should ovmenu. With Alt-F2, the user can choose whether to log in as xcsoar or root, but with the menu item, he has no such choice. (I prefer Alt-F2, but since I'll never fly with an OpenVario, my opinion doesn't count.)

kedder: > XCSoar shouldn't be running as root, and neither should ovmenu

I convinced myself that for embedded device like OV running as root is fine. Security considerations are not so critical: the device is not used to store sensitive information. And ngmenu (and ovshell for that matter) takes advantage of running as root by allowing to change some system configuration via their UIs (or, say, run opkg to receive package updates). Both ngmenu and ovshell also modifies xcsoar configuration, read flight logs, etc. So running menu and xcsoar as different users will pose additional challenges of working around the filesystem security obstacles.

Not saying it can't be done, but the benefits are not so apparent to me.

As for "exit to shell" - I'd try preserve the old behavior, but have no strong opinion, really. Leaving it to your judgement.

PR #210, opened 2022-01-24
small-fixes

lordfolken:


MaxKellermann:
Good PR
.... but what kind of horrible mess is this `psplash-ovlinux-img.h` file??
https://github.com/Openvario/meta-openvario/blob/1e1c6a9a596e5b49e99fc3e24bb41e2e4581e8bc/recipes-core/psplash/files/psplash-ovlinux-img.h#L7-L10
What the ... this isn't source code, this is automatically generated. Generated files should be generated at build time, not committed to git.

recipes-kernel/linux/linux-mainline/usbnet.cfg
```diff
@@ -12,4 +12,7 @@ CONFIG_MT76x0U=m
 # CONFIG_RT2800USB_UNKNOWN is not set
 # CONFIG_RTL8187 is not set
 CONFIG_RTL8XXXU=m
+CONFIG_RTL8XXXU_UNTESTED=y
+CONFIG_R8188EU=m
+CONFIG_STAGING=y
```
kedder, line 17: Not sure if `CONFIG_STAGING=y` belongs to `usbnet.cfg`. Probably a more generic location like `openvario.cfg` is more appropriate. But otherwise looks good to me!

kedder (approved):

recipes-kernel/linux/linux-mainline/usbnet.cfg
MaxKellermann: Putting it in `openvario.cfg` means it is always on, and then nobody knows why it's on. I think it's okay to have it right next to the driver which needs it (which may mean we have multiple copies of `STAGING=y`). In any case, there should be a comment explaining why `STAGING` is enabled.
The real question is whether we really want to support 1. staging drivers, and 2. this "untested" driver option.
Why are both `CONFIG_R8188EU` and `CONFIG_RTL8XXXU_UNTESTED` enabled? I don't understand what they are about.

kedder: > I think it's okay to have it right next to the driver which needs it 

Yes, until some other driver (in, say `usbserial.cfg`) also start depending on it. Basically, after a year or two it will not be obvious that `CONFIG_STAGING` can be removed from here safely, even if `CONFIG_R8188EU` is no longer "staging" at that point.

I'm fine with the change as it is, just pointing out that `CONFIG_STAGING` has broader scope than just USB network drivers.

MaxKellermann: But .. what's your point, what do you propose? That `STAGING` should always be kept on because nobody will ever be able to safely disable it? I don't think that's a good idea; I think we shouldn't enable `STAGING` at all, and not provide any support for unstable drivers, because that will likely cause problems. (Those drivers which were "staging" and "untested" in Linux 5.10 are still so in Linux 5.16. I don't see any progress here.)
About your general problem, nobody knows whether it's safe to remove: that would indeed be a problem, but the solution is easy, see https://docs.yoctoproject.org/kernel-dev/common.html#validating-configuration

linuxianer99 (approved):

recipes-kernel/linux/linux-mainline/usbnet.cfg
lordfolken: So the "Untested" option may not be necessary. Will Test.
As to the driver. Its been in staging since 3.12 and is even supplied on debian-stable (as staging) driver. 
If i have problem's with it, i will remove it.

Issue #203, opened 2022-01-23
ov57-lvds Display doesn't turn off when selecting poweroff


lordfolken:
Build: https://github.com/Openvario/meta-openvario/commit/1e1c6a9a596e5b49e99fc3e24bb41e2e4581e8bc
Machine: ov57-lvds (DS1)

The display on my ov57-lvds doesn't turn off when powering off. It fades to a white washed out screen and the backlight stays on. 
I then have to pull the fuse (both fuse/switch), to effectively power-off the device.

iglesiasmarianod:
Same behaviour here. Perhaps #103 might shed some light.
@kedder patched it for Warrior, I don't know if there's a full fix yet.

MaxKellermann:
The patch is still there, but appears to have no effect now. It's back to random behavior.

MaxKellermann:
Oh, look at this: https://github.com/torvalds/linux/commit/6eefb79d6f5bc4086bd02c76f1072dd4a8d9d9f6
This was also merged in 5.10.4, but Yocto-Hardknott is still at 5.10.2. (I hate using outdated kernels.)

lordfolken:
I upgraded it manually to 5.10.4, but no change. powering off from ovmenu or poweroff from the shell.

MaxKellermann:
That's what I meant when I wrote "appears to have no effect now". Something else broke it.
Unfortunately, there are no relevant fixes for this driver in newer kernels.

lordfolken:
For completion sake i also tried 5.10.93. Still no change.

MaxKellermann:
Looking at the PWM driver patch, I believe it never had any effect. This is a timing problem. I think the driver miscalculates the `next_period`, and submits the "disable" command too early.

kedder:
I can assure you it did fix an issue at that time. Not hard to verify by loading the old image. It's something else now, but with seemingly similar symptoms.

MaxKellermann:
I have found the real problem - it's a rounding error which has always been there, which leads to miscalculation of `next_period`, so my initial theory is correct.

The patch removing the `else` branch really does not solve the problem. It may change the timings in random ways that appear to make the problem go away, but that's just that: random coincidence.

kedder:
How random though? I didn't experience any randomness after the fix. Before the fix the "randomness" was clearly corellated with the brightness, as described here: https://github.com/Openvario/meta-openvario/issues/103#issuecomment-681132576.

[...]

PR #215, opened 2022-01-25
linux-mainline: import patches fixing the pwm-sun4i backlight problem

MaxKellermann:
Closes https://github.com/Openvario/meta-openvario/issues/203


kedder (approved):
@MaxKellermann are you willing to submit this upstream?

MaxKellermann:
I did, and I put you on Cc. https://lore.kernel.org/lkml/20220125123429.3490883-3-max.kellermann@gmail.com/

lordfolken:
So this works nicely for me. MHOTY max. 

What i observe is that even when the device is off, there is still about 0.08 A of current draw.

linuxianer99 (approved):

PR #218, opened 2022-01-25
 linux-mainline: build a non-modular kernel

MaxKellermann:
This enlarges the kernel somewhat; it grows from 4024 kB to 5008 kB
(+984kB), but at the same time, the /lib/modules folder shrinks from
5128 kB to 660 kB (-4468kB), leaving just the "r8188eu" driver module
(because it cannot be built statically).

This way, the kernel can be updated without touching the rootfs.


linuxianer99:
@MaxKellermann 
I did the switch to master branch.
Can you please rebase on this branch ??
Or is there any possibility for me to modify the PR ?

MaxKellermann:
GH doesn't allow me to modify the desired target branch... but you can just merge manually. (I have never used the GH merge feature. I always pull to my local repository, review there, and push to GH.)

kedder:
Honestly, I don't understand why we have to load all possible drivers into memory on every boot. I don't understand what problem it solves. "kernel can be updated without touching the rootfs" doesn't sound like a solution to any known problem.

MaxKellermann:
> Honestly, I don't understand why we have to load all possible drivers into memory on every boot.

We're talking about 1 MB of additional code in one existing file, in return for not having 4.5 MB of modules in 134 files, while at the same time omitting kmod and the systemd service connected to it.

1 MB is affordable, it's 0.1% of the RAM in the machine.

Compare:
- the bash that executes ovmenu-ng.sh: 3.3 MB
- plus "dialog": 2.7 MB.
- ts_uinput: 2 MB
- the "agetty" for the serial console: 1.9 MB

How much RAM does your ovshell consume, and how many files does it need to open during startup?

> I don't understand what problem it solves. "kernel can be updated without touching the rootfs" doesn't sound like a solution to any known problem.

I do know this problem.
It makes kernel updates simpler and more robust. So much can go wrong when kernel and modules are on different partitions, one of which cannot be accessed by a non-Linux computer. With a static kernel, anybody can just drop a new uImage on the boot partition (VFAT, acessible by all operating systems).

mihu-ov:
Is there any effect on time required to boot OV?

kedder:
Dunno, upgrading a kernel has never been a problem and why changing something that is not broken?

Upgrading kernel by dropping a file is simply not a use case. You can upgrade the whole system by flashing an image, or you can do a rolling upgrade using packages.  

I can understand building a kernel with statically linked drivers for a device that has fixed hardware that will never change (like a fridge or some smart thermostat), but for OV it is not that case - devices are connected dynamically; it makes sense to load only those drivers that support the connected devices. Also, some devices (like `r8188eu`) will apparently be in modules anyway (and those users will have to follow a special procedure for upgrading a kernel?).

You may have some higher goal in mind that requires a static kernel to reach. If so, please share, because I don't understand it. This change alone looks pretty irrational to me otherwise.

MaxKellermann:
> Is there any effect on time required to boot OV?

I care much about boot time, so that's a good point, but the TL/DR is: it doesn't matter.

It needs to load an additional megabyte, that takes more time, but that is barely measurable. Maybe 100ms if you assume that a SD card delivers 10 MB/s.

Not having to load individual kernel modules later compensates that disadvantage - loading many small files is, of course, much slower than loading one big chunk.

Look at this:
```
root@ov-cubieboard:~# systemd-analyze 
Startup finished in 2.868s (kernel) + 8.803s (userspace) = 11.671s 
multi-user.target reached after 8.553s in userspace
```
11.6 seconds is waaaaaay too slow. This really needs to be optimized. But the time for physically loading the kernel image from the SD card doesn't matter much here.

What does matter is replacing the ovmenu-ng with a big Python script. That will add lots of delay to the boot process, and will consume dozens of megabytes of additional RAM.

> You can upgrade the whole system by flashing an image

Which currently loses all data, and you have no plan how to fix that.

> or you can do a rolling upgrade using packages.

Have you ever done that with an OpenVario? I doubt it. Because that cannot work, currently. Upgrading the kernel package will have no effect.
It will put a new `uImage` on the root partition, but it needs to be on to boot partition. That's why I think nobody has ever upgraded the OpenVario kernel with `opkg`.

> I can understand building a kernel with statically linked drivers for a device that has fixed hardware that will never change (like a fridge or some smart thermostat), but for OV it is not that case - devices are connected dynamically; it makes sense to load only those drivers that support the connected devices.

This argument only appears to make sense, but it's a mirage. Because the list of modules is not dynamic. Your OpenVario will only have the modules which were compiled when the uImage was compiled. So your list of modules is just as static as my static uImage.

Yes, there is a theoretic possibility that some nerd takes the OpenVario kernel recipe, extends the kernel config to enable more modules, recompiles the kernel recipe, and copies just some `.ko` files on his OpenVario. But that scenario is not only unrealistic, but also more complicated than just copying a new `uImage`. And it's harder to do and more fragile, because you need exactly that kernel version with exactly that configuration, or else you have an ABI breakage with random consequences, including data corruption.

> Also, some devices (like r8188eu) will apparently be in modules anyway (and those users will have to follow a special procedure for upgrading a kernel?).

Because this driver is known to be bad. It was retrieved from a Realtek code dump and is in very bad shape. Just look at its ToDo file, that sounds bad - I wouldn't touch that driver. It hasn't improved, not even in the latest 5.16 kernel. If I could decide, I would disable it.

kedder:
> Not having to load individual kernel modules later compensates that disadvantage - loading many small files is, of course, much slower than loading one big chunk.

I can buy this argument if it is really more than theoretical improvement. But again, most usb drivers would not be ever loaded on a given device.

> > You can upgrade the whole system by flashing an image
> 
> Which currently loses all data, and you have no plan how to fix that.

Backup data to usb stick, flash new image to new SD card, restore. That's what you typically do once before the season, not a huge deal if you ask me. As a bonus you have your old known-to-work system on old sd card in case something goes wrong. 

> 
> > or you can do a rolling upgrade using packages.
> 
> Have you ever done that with an OpenVario? I doubt it. Because that cannot work, currently. Upgrading the kernel package will have no effect. It will put a new `uImage` on the root partition, but it needs to be on to boot partition. That's why I think nobody has ever upgraded the OpenVario kernel with `opkg`.

This is true, currently kernel is the only thing that is not upgradable through `opkg`. That would be nice to fix of course, then it should be possible to upgrade everything in the same way. But again, there's no big pressure to have it, because kernel is not something that typically brings visible improvements. 


> This argument only appears to make sense, but it's a mirage. Because the list of modules is not dynamic. Your OpenVario will only have the modules which were compiled when the uImage was compiled. So your list of modules is just as static as my static uImage.

If that would be true, all desktop linuxes would ship with statically linked modules. They all ship with fixed number of modules, but not a static kernel. Does, say, raspbian ship with static kernel? Why wouldn't they just stuff all possible drivers into the kernel binary if that's really a better option?

I don't really have much more to defend modular kernel other than what I mentioned and the fact that it seems to be a common, well established practice.

MaxKellermann:
> If that would be true, all desktop linuxes would ship with statically linked modules. They all ship with fixed number of modules, but not a static kernel. Does, say, raspbian ship with static kernel? Why wouldn't they just stuff all possible drivers into the kernel binary if that's really a better option?

You repeat this straw man argument again. "Somebody else is doing it our way, they figured out it's the best option for them, so logically follows it must also be the best for OpenVario. QED." - no, that's another fallacy. Don't fool yourself.

It's not that simple. Also, there's no objective "best option". It's a matter of balancing reasons. There's nothing objectively wrong with having a modular kernel; it has advantages and disadvantages. Just as well are there advantages and disadvantages in using a static kernel. It is just my opinion that for OpenVario, a static kernel is better, and I presented reasons for why I think so. 
All you ever replied was "I don't understand, doesn't solve any problem, look at those people over there, they disagree with you, so you must be wrong".

I did not say that all desktop Linux distributions should ship with a modular kernel. And that's not my opinion. The arguments have different weights for general-purpose distributions. A general-purpose distribution has a huge variety of hardware it needs to support.

Look at Debian Bullseye - its base kernel image weighs 6.6 MB (OpenVario: 4 MB), but its modules weigh 294 MB (OpenVario: 5 MB).

Even for Raspbian - a general-purpose distribution, but with a very narrow target hardware - I wouldn't suggest a static kernel. Basic kernel image is 7.6 MB, modules 82 MB. That's still 16x the size of OpenVario's modules.

For OpenVario, in my judgement, it's a different story. OpenVario is not a general-purpose distribution. The hardware is so narrow, you even have different kernel builds for the same board with a different display (which I think is excessive). The only variable hardware any user can have is stuff attached via USB. That's those 5 MB of modules which translate to just 1 MB size increase in a static kernel. In my opinion, that doesn't justify having the complexity for carrying the module infrastructure around, and the advantages for a static kernel prevail.

You disagree, that's okay. All your arguments look like straw men, but that's also okay, because there's no objective truth anyway.

linuxianer99 (approved):

lordfolken:
* Inlight of that the cubieboard2 is dead hardware, and that we will soon have to support different boards...
* The flexiblity given in what can be attached and removed to the whole openvario system. (with the board, without board, joystick, different serial adapters, wifi, bluetooth, soundcards etc.)
I don't think that moving to a single binary kernel is really a benefit at this point in time.

MaxKellermann:
> * Inlight of that the cubieboard2 is dead hardware, and that we will soon have to support different boards...
> 
>     * The flexiblity given in what can be attached and removed to the whole openvario system. (with the board, without board, joystick, different serial adapters, wifi, bluetooth, soundcards etc.)
>       I don't think that moving to a single binary kernel is really a benefit at this point in time.

What does this mean? I don't understand your point.

What does that have to do with the Cubieboard 2 deprecation? The decision modular vs static is the same on all boards.

No matter if your kernel is modular or static: the OpenVario firmware image needs to ship all drivers that will ever be used by anybody. No difference between modular and static.

lordfolken:
Your only argument for this is that you can copy a single file into /boot. 
But unless we do a ram disk with the entire ecosystem in it, this upgrade path doesn't make sense. No one will just replace the kernel by hand. 
Also there will be people that will hook their phone to their glider in the field to do upgrades. They would be very glad not to have to download huge images for every small change. 

Also this is contradictory to your stance on systemd where you want everything dynamically loaded on a need basis.

MaxKellermann:
> Your only argument for this is that you can copy a single file into /boot. But unless we do a ram disk with the entire ecosystem in it, this upgrade path doesn't make sense.

I don't understand this at all. How is an initramfs related to the question static/modular kernel?

> No one will just replace the kernel by hand. Also there will be people that will hook their phone to their glider in the field to do upgrades. They would be very glad not to have to download huge images for every small change.

And ... a modular kernel solves this how?
A modular kernel is about twice as large as the static kernel. If anything, then the static kernel *reduces* the amount of data that needs to be downloaded in the field.

> Also this is contradictory to your stance on systemd where you want everything dynamically loaded on a need basis.

A static kernel is monolithic; loading a static kernel is (nearly) as fast as loading a modular kernel without the modules; once you load at least one module, the static kernel wins.

But you cannot compared this to the situation in userspace; userspace is not monolithic. Not loading services until you really need them makes everything faster than loading everything unconditionally. This is very unlike a monolithic kernel.

kedder (approved):
I'm personally not convinced monolithic kernel is a good idea. I also think that the fact that some drivers cannot be compiled into the monolithic kernel makes an argument for it even weaker.

That said, I don't think it is harmful and I don't really have any strong argument against it. I'm not here to impede the progress :)

Issue #201, opened 2022-01-23
psplash missing --angle switch


lordfolken:
# issue:
machine: ov57-lvds
OV starts, the uboot splash screen is 90 degree rotated (after setting it in ovmenu settings).
Then the bootsplash animation done with psplash isn't rotated (0)
The ovmenu appears and is rotated 90 degrees again.

# Things i tried
When starting ```psplash --angle 90```from the shell, the boot splash is being displayed at the correct angle of 90.

# How to fix
I found a patch https://patchwork.openembedded.org/patch/123181/ that seems to read to read from /etc/rotation. However that doesn't solve our problem, as we have a (afaik) a static ramdisk. 

So I would suggest to add a small shell script to the ramdisk, that reads /proc/cmdline and parses the fbcon variable. 

The fbcon=rotate:3 switch would cause the shell script to output 90 as the argument for psplash.

Note: 
If found this, and in the basic machconf file there seems to be a general "CAN_ROTATE" switch. I haven't really found where this is being interpreted, but I think its the official way to go, and should be added to our machconf.
https://docs.yoctoproject.org/current/singleindex.html#adding-a-formfactor-configuration-file

lordfolken:
hmm the /etc/rotation doesn't seem to work.

lordfolken:
So i had a look at the psplash-init script and modifed it:
```bash
#!/bin/sh 
### BEGIN INIT INFO
# Provides:             psplash
# Required-Start:
# Required-Stop:
# Default-Start:        S
# Default-Stop:
### END INIT INFO

. /etc/default/rcS
export PSPLASH_FIFO_DIR

if [ ! -e /dev/fb0 ]; then
    echo "Framebuffer /dev/fb0 not detected"
    echo "Boot splashscreen disabled"
    exit 0;
fi

read cmdline < /proc/cmdline
for x in $cmdline; do
        case $x in
        psplash=false)
		echo "boot splashscreen disabled" 
		exit 0;
                ;;
        esac
done

rotation=0
read cmdline < /proc/cmdline
for x in $cmdline; do 
       case $x in 
         fbcon=rotate:0)
             rotation=0
             ;;
         fbcon=rotate:3)
             rotation=90
             ;;
         fbcon=rotate:2)
             rotation=180
             ;;
         fbcon=rotate:1)
             rotation=270
             ;;
       esac
done

[ -d $PSPLASH_FIFO_DIR ] || mkdir -p $PSPLASH_FIFO_DIR
[...]

PR #219, opened 2022-01-26
psplash rotation and more

lordfolken:
implements #201

### DIFF
--- a/meta-ov/recipes-core/psplash/files/psplash-rotation
+++ b/meta-ov/recipes-core/psplash/files/psplash-rotation
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# reads rotation from cmdline convert it to degrees and writes
+
+read -r cmdline < /proc/cmdline
+for param in ${cmdline} ; do
+  case "${param}" in
+      fbcon=*)
+          fbcon="${param#fbcon=}"
+          ;;
+  esac
+done
+
+case "${fbcon}" in
+    "rotate:3")
+        rotation=90
+        ;;
+    "rotate:2")
+        rotation=180
+        ;;
+    "rotate:1")
+        rotation=270
+        ;;
+    *)
+        rotation=0
+        ;;
+esac
+exec /usr/bin/psplash --angle "${rotation}"
--- a/meta-ov/recipes-core/psplash/files/psplash-start.service
+++ b/meta-ov/recipes-core/psplash/files/psplash-start.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Start psplash boot splash screen
+DefaultDependencies=no
+RequiresMountsFor=/run
+
+[Service]
+Type=notify
+ExecStart=/opt/bin/psplash-rotation
+RemainAfterExit=yes
+
+[Install]
+WantedBy=sysinit.target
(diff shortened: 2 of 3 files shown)


meta-ov/recipes-core/psplash/files/rotation
```diff
@@ -0,0 +1 @@
+ROTATION=0
```
kedder, line 1: Hm, we have this rotation in bunch of files already. I wonder if we can use info from /boot or /dev (don't have exact paths at hand at the moment, but I can look up if necessary.

meta-ov/recipes-core/psplash/files/psplash-start.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Start psplash boot splash screen
+DefaultDependencies=no
+RequiresMountsFor=/run
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/psplash --angle "${ROTATION}"
+EnvironmentFile=/etc/rotation
```
kedder, line 9: Another solution that comes to mind is to run some shell script instead of /usr/bin/psplash directly, where we can find the correct rotation and pass it as `--angle` parameter.

lordfolken: I wasted hours on this... The problem is that psplash-start communicates with psplash-systemd via a socket. (Hence type notify) it also expects it to remain after exit.
A shell wrapper destroys that functionality and even leaves ov unbootable. 

The only way to pass an env variable inside a systemd unit is via Environment or as above EnvironmentFile. Environmentfile has to exist before any process is started. So you cannot have a ExecStartPre process that creates a variable file. 

So the only path left is to write another unit that runs before psplash and wites the environment file. (i'm doing that now.)

MaxKellermann: > I wasted hours on this... The problem is that psplash-start communicates with psplash-systemd via a socket. (Hence type notify) it also expects it to remain after exit. A shell wrapper destroys that functionality and even leaves ov unbootable.

Not if you `exec` the final program.

MaxKellermann:
Why are there two distinct systemd units for this? This looks unnecessarily complicated.

lordfolken:
> Why are there two distinct systemd units for this? This looks unnecessarily complicated.

There is no way to hand over a dynamically assinged variable inside a systemd unit (thx leonard):
* The environmentfile has to exist before invocation of the systemd unit, and is also read first. So ExecStartPre is also not an option. 
* As psplash can' t read a configfile it has to be a variable. 
* As psplash requires socket notification (type notify) in order to display progress, it cannot be shell wrapped.

MaxKellermann:
>     * As psplash requires socket notification (type notify) in order to display progress, it cannot be shell wrapped.

Yes it can. See https://github.com/Openvario/meta-openvario/pull/219#discussion_r793270479

meta-ov/recipes-core/psplash/files/psplash-rotation
```diff
@@ -0,0 +1,30 @@
...
+    "rotate:2")
+        echo "ROTATION=180" > "${rotation}"
+        ;;
+    "rotate:1")
+        echo "ROTATION=270" > "${rotation}"
+        ;;
+    *)
+        echo "ROTATION=0" > "${rotation}"
+        ;;
+esac
```
kedder, line 30: I suspect (as @MaxKellermann suggests), that if you add this line here it will run just fine. Can you confirm / deny this?

```
exec /usr/bin/psplash --angle 90
```

lordfolken: Yes it does. So the new version contains just that. I still find shell wrappers in systemd units just yuck.

kedder (approved):
Looks great now, thanks!

linuxianer99 (approved):

PR #220, opened 2022-01-27
ovscripts: reimplement download to usb stick script

lordfolken:
implements #225


meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
```diff
@@ -1,19 +1,18 @@
 #!/bin/bash
 
-USB_PATH="/usb/usbstick/openvario/download/xcsoar"
+USB_PATH="/usb/usbstick/openvario/download/xcsoar/"
 DOWNLOAD_PATH="/home/root/.xcsoar"
 
-if [ ! -d "$USB_PATH" ]; then
-	mkdir -p "$USB_PATH"
+if [ ! -d "${USB_PATH}" ]; then
+	mkdir -p "${USB_PATH}"
```
MaxKellermann, line 7: What's the point of that?

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
```diff
@@ -1,19 +1,18 @@
...
 fi
 
-if [ -z "$(ls $DOWNLOAD_PATH/* 2>/dev/null)" ]; then
+if [ -z "$(ls ${DOWNLOAD_PATH}/* 2>/dev/null)" ]; then
```
MaxKellermann, line 10: And that? No explanation in the commit message what exactly is being fixed ...

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-all.sh
```diff
@@ -2,13 +2,10 @@
 
 USB_PATH="/usb/usbstick/openvario/upload"
 UPLOAD_PATH="/home/root/.xcsoar"
-if [ -z "$(ls $USB_PATH/* 2>/dev/null)" ]; then
+if [ -z "$(ls ${USB_PATH}/* 2>/dev/null)" ]; then
```
MaxKellermann, line 5: You add curly braces (why??), but the parameter is still unquoted; everywhere else, expanded variables are quoted.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-all.sh
```diff
@@ -2,13 +2,10 @@
...
         echo "No files found !!!"
 else
-        for uploadfile in $(find $USB_PATH -type f); do
-                echo $uploadfile
-                cp $uploadfile $UPLOAD_PATH
-        done
+        cp -rv "${USB_PATH}/." "${UPLOAD_PATH}"
```
MaxKellermann, line 8: Why the `-v`? It wasn't there before, is it part of fixing something?

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
lordfolken: curly braces are best practice.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-all.sh
lordfolken: previously each file was echoed to the gui box that ovmenu-ng invokes, to display what was copied to the user. This emulates that echo with a single command.

lordfolken: ovmenu.ng expects the output of the program to be verbose. since we are no longer echoing, -v will add output which is displayed to the user.

lordfolken:
There is an error with recursion.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
```diff
@@ -7,13 +7,12 @@ if [ ! -d "$USB_PATH" ]; then
 	mkdir -p "$USB_PATH"
 fi
 
-if [ -z "$(ls $DOWNLOAD_PATH/* 2>/dev/null)" ]; then
+if [ -z "$(find "$DOWNLOAD_PATH" -type f 2>/dev/null)" ]; then
```
MaxKellermann, line 10: Insert a `|head -1` so `find` gets killed with `SIGPIPE` after finding the first file; with huge directory trees, that will be a big speedup.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
```diff
@@ -7,13 +7,12 @@ if [ ! -d "$USB_PATH" ]; then
...
         echo "No files found !!!"
 else
-        cp -rv "$DOWNLOAD_PATH/*" "$USB_PATH"
+        find "$DOWNLOAD_PATH" -type f -exec basename {} \; -exec cp "{}" "$USB_PATH" \;
```
MaxKellermann, line 13: Keeping the `-v` would be easier and faster than invoking `basename`

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-xcsoar.sh
```diff
@@ -2,15 +2,11 @@
...
         echo "No files found !!!"
 else
-        for uploadfile in $(find $USB_PATH -type f); do
-                echo $uploadfile
-                cp $uploadfile $UPLOAD_PATH
-        done
+        find "$USB_PATH" -type f -exec basename {} \; -exec cp "{}" "$UPLOAD_PATH" \;
 fi
 echo "Umount Stick ..."
-sync
```
MaxKellermann, line 14: Why remove the `sync`?

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
lordfolken: good point. Will do.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-xcsoar.sh
August2111: Because of the unmount, the sync should be automatic - but my question is: Why is the unmount happening here? The shell script does not mount the USB stick, with subsequent actions (e.g. download-all.sh) the stick is unexpectedly not mounted! The same applies to all other scripts!
From my point of view, the 'unmount' should be its own menu item!

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-xcsoar.sh
```diff
@@ -2,15 +2,11 @@
 
 USB_PATH="/usb/usbstick/openvario/upload/xcsoar"
```
August2111, line 3: BTW In newer versions the USB stick is mounted as "/usb" only - and not as "/usb/usbstick" like in previous versions

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-xcsoar.sh
MaxKellermann: > Because of the unmount, the sync should be automatic

Only sync to the to-be-unmounted filesystem is implicit. Which is not relevant here.

If removing a `sync` command is useful, that fact should be documented, and should not be hidden inside a unrelated commit. That makes patch review harder and can lead to unintended regressions.

> - but my question is: Why is the unmount happening here? The shell script does not mount the USB stick

That is a good question, and I was wondering the same - but since this is old code which is left untouched in this PR, I didn't ask it here.

August2111: @MaxKellermann: Are you using the scripts in your new OV menu in XCSoar? Or is the effort here obsolete?
In addition, we also have Kedders (Python) menu...

MaxKellermann: Yes, so cleanup here is good - even if they get ported to C++ eventually, the lessons learned from existing code is always useful.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-xcsoar.sh
lordfolken: this was an error #227 and corrected since.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
lordfolken: yes but at least on my 5.7" in portrait mode, that lists the full path and file name is just scrolled off screen.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/download-all.sh
lordfolken: done

meta-ov/recipes-apps/ovmenu-ng-skripts/files/upload-xcsoar.sh
lordfolken: I re added the sync in my new version, but removed the umount. This worked before, as it was immediately remounted by autofs.

linuxianer99 (approved):
Looks good to me

MaxKellermann:
Inconsistent indentation; in BitBake files and shell scripts, we currently indent with tabs. This was an inconsistent mess previously, and I submitted lots of fixups for that. This PR reintroduces new inconsistencies.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
MaxKellermann, line 1: Can we do without bash? I think this script is simple enough, and eliminating bash would allow us to cut 1 MB from the image.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
```diff
@@ -0,0 +1,47 @@
...
+  'upload-xcsoar.sh')
+    SRC_PATH="$USB_PATH/upload/xcsoar"
+    DEST_PATH="$XCSOAR_PATH"
+    ;;
+  'upload-all.sh')
+    SRC_PATH="$USB_PATH/upload"
+    DEST_PATH="$XCSOAR_PATH"
+    ;;
+  *)
+    echo 'call as download-all.sh, upload-xcsoar.sh or upload-all.sh'
```
MaxKellermann, line 22: Error messages must go to stderr, not stdout

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
```diff
@@ -0,0 +1,47 @@
...
+  exit 1
+fi
+
+if [ -z "$(find "$SRC_PATH" -type f | head -n 1 2>/dev/null)" ]; then
+  echo 'No files found !!!'
+else
+  # echo and copy each file / directory for tui
+  cd "$SRC_PATH" || exit 1
+  find ./ -type d -exec echo {} \; -exec mkdir -p "$DEST_PATH/{}" \;
+  if find ./ -type f -exec echo {} \; -exec cp "{}" "$DEST_PATH/{}" \; ; then
```
MaxKellermann, line 37: Instead of checks like that, I think it would be simpler to just use `set -e`

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
linuxianer99: i am ok without bash.

lordfolken:
> Inconsistent indentation; in BitBake files and shell scripts, we currently indent with tabs. This was an inconsistent mess previously, and I submitted lots of fixups for that. This PR reintroduces new inconsistencies.
Converting to spaces would be a much saner choice.
Well then converted for "consistency".

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
lordfolken: Yes, but then it would be hard to display an end-user readable error message.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
lordfolken: download-igc.sh presently depends on bash. but that thing is anyway in need of a rework.

lordfolken: this script will run on sh just fine. converted.

MaxKellermann:
I don't care much about tabs vs spaces, but there should be a decision on either, and it should be used consistently. From existing code, it looked like tab prevailed, so I made everything I did with tabs, and fixed lots of mixed files.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
MaxKellermann: I thought the error message printed by `cp` was already good enough.
But I just found out that `find` does not propagate `-exec` failures to the caller. It just keeps on executing more commands even if one has failed, and exits with status 0 (success). Did you know that? (I didn't because I never use `find -exec`)

lordfolken: It aborts on first error, yet the return code is 0. What a dangerous design. 
```shell
touch 1 2 3
find -exec cat "{}" \;
cat: ./: Is a directory
echo $?
0
```
So my if check statement doesn't catch what i thought it would.

There is however the -exec {} +  form, that one would return a non-zero value from an execd process. 

So in the end this too much complexity. I will redesign.

lordfolken: So my new approach uses rsync. I know 499kb + , but:
The transfers to the stick now take mere seconds (even with checksumming) for a 400MB takes a mere 20seconds (with a previous copy on the stick) as compared for the about 20minutes with the previous solution.
I took care to use -c for checksum comparision  only, ignoring filesize and timestamps.

linuxianer99:
> I don't care much about tabs vs spaces, but there should be a decision on either, and it should be used consistently. From existing code, it looked like tab prevailed, so I made everything I did with tabs, and fixed lots of mixed files.

I have also no preference regarding tabs or spaces. But as Max already started with tabs now, lets use tabs.

lordfolken:
> > I don't care much about tabs vs spaces, but there should be a decision on either, and it should be used consistently. From existing code, it looked like tab prevailed, so I made everything I did with tabs, and fixed lots of mixed files.
> 
> I have also no preference regarding tabs or spaces. But as Max already started with tabs now, lets use tabs.

My push is now tabbed.

meta-ov/recipes-apps/ovmenu-ng-skripts/files/transfer-xcsoar.sh
```diff
@@ -0,0 +1,45 @@
...
+	'upload-xcsoar.sh')
+		SRC_PATH="$USB_PATH/upload/xcsoar"
+		DEST_PATH="$XCSOAR_PATH"
+		;;
+	'upload-all.sh')
+		SRC_PATH="$USB_PATH/upload"
+		DEST_PATH="$XCSOAR_PATH"
+		;;
+	*)
+		>&2 echo 'call as download-all.sh, upload-xcsoar.sh or upload-all.sh'
```
MaxKellermann, line 22: Strange way .. of placing the `>&2` but perfectly legal.

meta-ov/recipes-apps/ovmenu-ng-skripts/ovmenu-ng-skripts_0.1.bb
```diff
@@ -41,11 +40,13 @@ do_install() {
...
-		${S}/upload-all.sh \
-		${S}/upload-xcsoar.sh \
-		${S}/download-all.sh \
+		${S}/transfer-xcsoar.sh \
 		${S}/ov-calibrate-ts.sh \
 		${D}${bindir}/
+        cd ${D}${bindir}
+        ln -s -r transfer-xcsoar.sh upload-all.sh
+        ln -s -r transfer-xcsoar.sh upload-xcsoar.sh
+        ln -s -r transfer-xcsoar.sh download-all.sh
```
MaxKellermann: Spaces here instead ot ab

meta-ov/recipes-apps/ovmenu-ng-skripts/ovmenu-ng-skripts_0.1.bb
```diff
@@ -30,23 +30,23 @@ SRC_URI =      "\
 addtask do_package_write_ipk after do_package
 
 do_compile() {
-        :
+				:
```
MaxKellermann, line 33: 4 tabs??

meta-ov/recipes-apps/ovmenu-ng-skripts/ovmenu-ng-skripts_0.1.bb
```diff
@@ -30,23 +30,23 @@ SRC_URI =      "\
...
 }
 
 do_install() {
-        echo "Installing ..."
-        install -d ${D}${bindir}
-        install -m 0755 \
+				echo "Installing ..."
```
MaxKellermann, line 37: 4 tabs??

meta-ov/recipes-apps/ovmenu-ng-skripts/ovmenu-ng-skripts_0.1.bb
```diff
@@ -30,23 +30,23 @@ SRC_URI =      "\
...
 		${S}/update-system.sh \
 		${S}/download-igc.sh \
 		${S}/transfer-xcsoar.sh \
 		${S}/ov-calibrate-ts.sh \
 		${D}${bindir}/
-        cd ${D}${bindir}
-        ln -s -r transfer-xcsoar.sh upload-all.sh
-        ln -s -r transfer-xcsoar.sh upload-xcsoar.sh
-        ln -s -r transfer-xcsoar.sh download-all.sh
+				cd ${D}${bindir}
```
MaxKellermann, line 46: 4 tabs??

MaxKellermann (changes requested):
Indentation is still a bit jumbled

lordfolken:
> Indentation is still a bit jumbled

I reindented and retabbed.

linuxianer99:
So ... are we ready to merge ??
From my point of view, it is ok now.

lordfolken:
Wait. Got mixed up

linuxianer99:
> Wait. Got mixed up

ok ..

lordfolken:
> > Wait. Got mixed up
> 
> ok ..
so now.

PR #222, opened 2022-01-27
Kernel config refactoring

MaxKellermann:
Addresses https://github.com/Openvario/meta-openvario/discussions/185#discussioncomment-2005345 (Dallas 1-wire)
This is a variant of https://github.com/Openvario/meta-openvario/pull/218 which omits the commit which makes everything static, because there was not yet a general agreement.


linuxianer99 (approved):
Looks good to me

meta-ov/recipes-kernel/linux/linux-mainline/openvario.cfg
```diff
@@ -27,7 +27,7 @@ CONFIG_UEVENT_HELPER=y
 
 # CONFIG_INPUT_FF_MEMLESS is not set
 # CONFIG_INPUT_MOUSE is not set
-CONFIG_INPUT_UINPUT=m
+CONFIG_INPUT_UINPUT=y
```
kedder, line 30: > It is needed by ts_uinput, so it's always loaded, and there's no point in making it modular.

Well, unless device has no touchscreen. But I guess we are running `ts_uinput` regardless, so in practice it is getting loaded all the time anyways.

kedder (approved):
Looks good!

PR #231, opened 2022-01-28
psplash: update override syntax to yocto 3.4 standard

lordfolken:
The hornister update requires a newer syntax. (and we need a github action that verifies bitbake recipes at least.)

### DIFF
--- a/meta-ov/recipes-core/psplash/psplash_git.bbappend
+++ b/meta-ov/recipes-core/psplash/psplash_git.bbappend
@@ -5,9 +5,9 @@ SPLASH_IMAGES = "file://psplash-ovlinux-img.h;outsuffix=default"
 SRC_URI += "file://psplash-rotation \
             file://psplash-start.service"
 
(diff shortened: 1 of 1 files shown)


CazYokoyama:
"bitbake openvario-image" becomes buildable

kedder (approved):

linuxianer99:
> and we need a github action that verifies bitbake recipes at least.

@lordfolken  Do you have an idea, how to verify ??
I am working on automated build using github actions at the moment, but the build take about 2h per target ....

linuxianer99 (approved):

lordfolken:
> > and we need a github action that verifies bitbake recipes at least.
> 
> @lordfolken Do you have an idea, how to verify ?? I am working on automated build using github actions at the moment, but the build take about 2h per target ....

@linuxianer99 There seems to be a --dry-run switch to bitbake. That parses but doesn't build.

PR #233, opened 2022-01-29
Create CI Action to verify recipes at PRs

linuxianer99:
This creates a Github Action that run a bitbake dry-run to verify the recipe syntax

### DIFF
--- a/.github/workflows/verify_p.yml
+++ b/.github/workflows/verify_p.yml
@@ -21,7 +21,13 @@ jobs:
     
     # The type of runner that the job will run on
     runs-on: ubuntu-20.04
-    
+
+    # Matrix Run: Test all combinations
+    strategy:
+      matrix:
+        MACHINE: [openvario-57-lvds, openvario-7-PQ070, openvario-7-CH070]
+        TARGET: [openvario-image-testing, openvario-image]
+
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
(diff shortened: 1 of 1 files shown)


linuxianer99:
@lordfolken  is this what you have in mind ?

lordfolken:
Yes something like that should do it.
Two things come to mind:
 * use the hcr.io/openvario/ovbuild-container
 * verify all machine types: Github has this [matrix](https://ncorti.com/blog/howto-github-actions-build-matrix) function.

.github/workflows/verify_p.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/checkout@v2
+        
+      - name: Update submodules
+        run: git submodule update --init
+      
+      # Install necessary packages
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ffmpeg gawk wget git diffstat unzip texinfo gcc build-essential chrpath socat cpio python3 python3-pip python3-pexpect xz-utils debianutils iputils-ping python3-git python3-jinja2 libegl1-mesa libsdl1.2-dev pylint3 xterm python3-subunit mesa-common-dev zstd liblz4-tool
```
MaxKellermann, line 39: What's the use of these packages? They are only needed to actually build XCSoar.

linuxianer99: Yes. I am also working on a different container for building complete images, not only checking recipes. I try to minimize the variants of the workflow files. 
For only checking the recipes (--dry-run) these packages are not necessary.

linuxianer99:
> Yes something like that should do it. Two things come to mind:
> 
> * use the hcr.io/openvario/ovbuild-container
> * verify all machine types: Github has this [matrix](https://ncorti.com/blog/howto-github-actions-build-matrix) function.

@lordfolken i will have a look on this ... but i think for the recipe check, it doesn't matter, which image or target is built ....

linuxianer99:
@lordfolken I decided not to use the container, as it is unnecessary overhead in a defined build environment like the github actions.

Matrix build is included ..

PR #236, opened 2022-02-02
Optimize runtime of github action

linuxianer99:
Build both targets using one bitbake command


linuxianer99:
@MaxKellermann  Like this ?

kedder (approved):

MaxKellermann:
Yes. Much better.

MaxKellermann:
I would even run all MACHINEs in one job. The matrix is a nice feature if you want to check different build machines, but Yocto builds are almost completely decoupled from the build machine (different OS, different CPU architecture, compiler version etc.). The dry-run will also not benefit much from parallelizing the checks for different MACHINEs.

Issue #232, opened 2022-01-29
language at XCSoar is always English


Blaubart:
I'm not able to change language at XCSoar using the OpenVario menu.

bomilkar:
Make sure LANG is set before xcsoar starts.
For instance "LANG=de_DE.UTF-8".
This is a bad workaround but it may help pinning down the root cause.

MaxKellermann:
This is because `ovmenu-ng.sh` is bugged:
https://github.com/Openvario/meta-openvario/blob/42b5de934d3d95f76ca514c069cac90d01f1145b/meta-ov/recipes-apps/ovmenu-ng/files/ovmenu-ng.sh#L13
On my OpenVario, this expands to:
```
source /opt/conf/machine.conf /opt/conf/ov-xcsoar.conf /opt/conf/sensord.conf /opt/conf/variod.conf
```
The author expected this to read all `*.conf` files, but it does not.
Look at https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bourne-Shell-Builtins
> . filename [arguments]

This will only read the first file (i.e. `machine.conf`) and pass the names of all other files as command-line arguments to it.

Fortunately, it did not execute all files ... because `sensord.conf` and `variod.conf` files are not shell scripts. Executing them spits some error messages at best.

PR #237, opened 2022-02-02
Recipe cleanups

MaxKellermann:


MaxKellermann:
I've added more commits to get rid of the useless and bugged `xcsoar_config.sh` script. See https://github.com/Openvario/meta-openvario/discussions/234#discussioncomment-2096291

MaxKellermann:
I'll add another patch to address https://github.com/Openvario/meta-openvario/issues/232

MaxKellermann:
I've added two more patches which simplify the initial u-boot environment by turning the big C patch file into a small text file. Turns this mess:
https://github.com/Openvario/meta-openvario/blob/42b5de934d3d95f76ca514c069cac90d01f1145b/meta-ov/recipes-bsp/u-boot/files/0001-Environment-Openvario-mainline.patch#L64-L71
into:
https://github.com/Openvario/meta-openvario/blob/aa342dc6bc1bb22fcad9b8440ab25e6f762bceac/meta-ov/recipes-bsp/u-boot/files/bootenv.ini#L34-L49
This PR is becoming slightly off-topic; I can split it if you prefer picking only some parts now.

PR #244, opened 2022-02-03
Various ov-menu recipe fixups

MaxKellermann:


MaxKellermann:
In particular, the "stderr to journal" commit fixes a bug that freezes XCSoar after a while - this is important! Was this never noticed before by anybody else?

linuxianer99 (approved):

kedder (approved):
LGTM

lordfolken:
> In particular, the "stderr to journal" commit fixes a bug that freezes XCSoar after a while - this is important! Was this never noticed before by anybody else?

i have run xcsoar on ov continously for several days without issues.

PR #249, opened 2022-02-04
gitignore: add build dirs/artefacts and bitbake.lock

lordfolken:
git status should be clean after a build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,10 @@
 *~
 *.log
+OpenVario*rootfs.img.gz
 .DS_Store
+bitbake.lock
 cache/
 downloads/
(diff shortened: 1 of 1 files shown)


linuxianer99:
@lordfolken This is only valid, if the build directory is initialized within the source folder ...

like 

`source openembedded-core/oe-init-build-env .`

if you point it to a different directory, it makes no sense ...
But for me it is ok, to merge ...

lordfolken:
Which are currently the instructions in the readme. :-)

linuxianer99:
> Which are currently the instructions in the readme. :-)

That's right  :-)

Issue #247, opened 2022-02-03
openvario-recovery-image doesn't build


lordfolken:
current master, doesn't build the openvario-recovery-image target.
The bibtake validation already fails, with checksum errors in the licence files. 
Once that is fixed, it fails to build the openvario-recover-image initramfs.

The target openvario-recovery-image should be added to the validation workflow @linuxianer99

lordfolken:
```shell
WARNING: openvario-recovery-image-1.0-r24 do_populate_lic: ${COREBASE}/LICENSE is not a valid license file, please use '${COMMON_LICENSE_DIR}/MIT' for a MIT License file in LIC_FILES_CHKSUM. This will become an error in the future
ERROR: openvario-recovery-image-1.0-r24 do_populate_lic: QA Issue: openvario-recovery-image: The LIC_FILES_CHKSUM does not match for file:///workdir/openembedded-core/LICENSE;md5=4d92cd373abda3937c2bc47fbc49d690
openvario-recovery-image: The new md5 checksum is b97a012949927931feb7793eee5ed924
openvario-recovery-image: Here is the selected license text:```

PR #250, opened 2022-02-04
Add openvario-recovery-image

linuxianer99:


linuxianer99:
@lordfolken  Like this ?

lordfolken:
as stated in #247 
yes thanks. Interestingly enough this fails on my machine, but succeeds here?

PR #257, opened 2022-02-05
xcsoar: remove obsolete dlgCredits patch

lordfolken:
My [PR](https://github.com/XCSoar/XCSoar/pull/862/) merged into XCSoar today breaks this patch. The PR also obsoletes the patch.


linuxianer99:
Should we merge this, if the new XCSoar is released ??
Or can we merge this now ?

lordfolken:
I would merge now, as currently bitbake will not build openvario-image-tesitng without this commit. 
openvario-image doesn't run anayway and the patch is in master for 7.23.

MaxKellermann:
It could be disabled only from the -testing recipe now (via `:remove`), but since this patch is rather unimportant, I'd say this distinction isn't worth it. Merge now.

Issue #256, opened 2022-02-05
rotation is now bonkers


lordfolken:
Setting the rotation in ovmenu sets the rotation correct at that instant. If you reboot however both psplash and fb are rotated differently.

lordfolken:
its also only rotates the first console, not alt-f2.

lordfolken:
root@openvario-57-lvds:~# cat 3.prereboot 3.postreboot 
rotation=3
console=ttyS0,115200 root=/dev/mmcblk0p2 rootwait panic=10 lima.job_hang_limit=10 fbcon=rotate:2 vconsole.font=default8x16
3
rotation=3
console=ttyS0,115200 root=/dev/mmcblk0p2 rootwait panic=10 lima.job_hang_limit=10 fbcon=rotate:1 vconsole.font=default8x16
1```bash

lordfolken:
so basically when rotation=1 console should 3. and vica versa.

MaxKellermann:
Yes, I forgot to apply this translation:
https://github.com/Openvario/meta-openvario/blob/5f85d706c1744835db145129acc82c53a0bbbd61/meta-ov/recipes-bsp/u-boot/files/bootenv.ini#L37-L41
Can somebody explain to me why this translation exists? Why does OpenVario redefine the meaning of these rotation parameters, why are we not simply using the kernel's definition? This only complicates things, but does not appear to make sense. @linuxianer99, do you know?

lordfolken:
I found this now also after searching. I'm preparing a PR at the moment for this.

lordfolken:
I still have a problem with psplash. The /sys/class/graphics/fbcon/rotate returns 0 at boot, even if the consoles et all are rotated by the kernel argument.

lordfolken:
I suspect currently this kernel option:
CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=n

lordfolken:
setting  CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER to n fixes the problem with psplash. (but displays penguins.)

MaxKellermann:
Indeed a surprising artefact - but why is it a problem?

linuxianer99:
> Yes, I forgot to apply this translation:
> 
[...]

PR #259, opened 2022-02-06
Fixrotation

lordfolken:


MaxKellermann:
The second commit will break existing configurations

linuxianer99:
ok, we wait with mergeing until #256 is fixed ?

MaxKellermann:
This is supposed to be the fix for #256, but it solves it in a way that breaks compatibility with existing configurations.
You could "don't care", then everybody "just" has to reconfigure the rotation, and we'd get rid of some cruft complexity, that's good. Both is ok for me. It just needs to be an explicit decision, and not some "oops we broke it".

linuxianer99:
ok... i see ..
I will merge, as we have a lot of huge changes at the moment ... so this will be no problem :-)

PR #260, opened 2022-02-07
psplash: rotation based on /boot/config.uEnv

lordfolken:


MaxKellermann:
This is better than `/proc/cmdline`, because it reflects the effective currently desired rotation; for example, if we some day use psplash at runtime (for whatever theoretical reason - this example is just here to demonstrate the robustness of the solution), and the user has reconfigured rotation, but hasn't rebooted, `/proc/cmdline` will not reflect the new value.
However, your diff is occluded by the indentation changes you did in the same patch. What could be a trivial one-line change has become a diff which changes almost all lines.

Issue #248, opened 2022-02-04
uboot splash screen no longer displays


lordfolken:
I no longer have a uboot splash screen on openvario-lvds57. It remains black, the backlight turns on and at some point psplash takes over.

PR #261, opened 2022-02-08
Ubootsplash

lordfolken:
This fixes the uboot display of the boot image. #248


meta-ov/recipes-core/psplash/files/psplash-start.service
```diff
@@ -2,6 +2,7 @@
 Description=Start psplash boot splash screen
 DefaultDependencies=no
 RequiresMountsFor=/run
+RequiresMountsFor=/boot
```
MaxKellermann, line 5: btw. this will delay the whole boot process until /boot is mounted. It is a dependency I wanted to avoid; this means we can remove the automount feature.

lordfolken: Nov 10 12:58:34 openvario-57-lvds kernel: Booting Linux on physical CPU 0x0
Nov 10 12:58:36 openvario-57-lvds systemd[1]: Mounting Mount the boot partition...
Nov 10 12:58:37 openvario-57-lvds systemd[1]: Started Start psplash boot splash screen.
Nov 10 12:58:41 openvario-57-lvds systemd[1]: Started OVMenu Next-Generation.
Nov 10 12:58:43 openvario-57-lvds psplash-systemd[153]: Systemd reported progress of 1.0, quit psplash.

There is no noticeable difference here.
We are loosing 10seconds between uboot and psplash, that worries me more.

PR #266, opened 2022-02-09
SVG logo

MaxKellermann:
For now, this is just for u-boot, but I'll integrate this in psplash as well, and of course in the new [XCSoar-based graphical menu](https://github.com/Openvario/meta-openvario/pull/199).

New look:
new

Old look:
old

Thanks @MacflySF27 and his daughter.


linuxianer99 (approved):

MaxKellermann:
I added two more patches; the first implements BMP-RLE support for the Cubieboard 2 (which has a 32 bit display, and u-boot implements BMP-RLE only for 16 bit displays); the second patch enables RLE compression, reducing the /boot partition (and the image size) by several megabytes.

mihu-ov:
Looks good! Still the logo we are used to but with a refreshed look.

MaxKellermann:
The look previously was rather ugly; not because the logo design was bad, but because it was a scaled bitmap, and thus was very blurry. Now all images get renderered at their native resolution.

(And, of course, all images get rendered automatically during the build. Nobody needs to do that manually. This PR deletes 56 logo files which were previously managed manually!)

linuxianer99:
@kedder Ok to merrge ?

meta-ov/recipes-bsp/u-boot/files/0001-video_bmp-implement-BMP-RLE-to-32-bit.patch
```diff
@@ -0,0 +1,160 @@
+From fdad291b7837db232a9a8d071d8d587d8777a485 Mon Sep 17 00:00:00 2001
+From: Max Kellermann <max.kellermann@gmail.com>
+Date: Wed, 9 Feb 2022 17:55:27 +0100
+Subject: [PATCH] video_bmp: implement BMP-RLE to 32 bit
```
kedder, line 4: Are you planning to submit this upstream? It might be a PITA to fix this large patch once it stops applying cleanly.

kedder (approved):

PR #267, opened 2022-02-10
Display u-boot slash early

MaxKellermann:
Lots of u-boot optimizations, but the most important piece is "display OpenVario logo in preboot", which loads and displays the splash screen before scanning for USB devices and before the 2 second delay (waiting for "ov" to interrupt the boot process). This shows the splash screen *less than a second* after power-up.


mihu-ov:
Should we set bootdelay to -2 to boot without delay and not check for abort (see https://www.denx.de/wiki/view/DULG/UBootEnvVariables)?  Or is there any need to check for abort other than for development?
IIRC if a device sends any data on the serial port during these 2 seconds this may stop the boot process.

MaxKellermann:
I had the same idea, but wasn't sure why you have this 2 second delay. It is configured explicitly in u-boot, even with a custom key code, i.e. it's not left enabled by accident.
If I were to decide, I would disable this delay completely and instead let the Cubieboard2 on-board button interrupt the boot process. If you need to debug your OpenVario, I think it's okay to have to open the case to reach that button. Normal users will never debug on the u-boot command line (will anybody ever?), and normal users will enjoy 2 seconds faster booting.

Another 5 seconds of boot time is lost while scanning for USB devices. Such a long delay is very annoying, but this is necessary for USB stick recovery. I wish there was another way...

linuxianer99:
The 2 second delay and the special key code for interrupting the boot was intentionally. 
But i agree fully with you, that nobody will debug at uboot, despite for the recovery boot (and this is a very rare case)

i agree with the proposal of Max:

- Remove the boot delay completely
- Use a button for interrupting the boot process

mihu-ov:
Re scanning of USB devices, slide 51 from https://elinux.org/images/9/99/Opdenacker-timing-boot-time-reduction-techniques-ELCE-2019.pdf identifies the same issue, "_Waiting for 1.2 s for the USB camera to be enumerated. Any way around this USB asynchronousness issue_?"
However, nothing is mentioned in (newer) https://bootlin.com/doc/training/boot-time/boot-time-slides.pdf

MaxKellermann:
> Another 5 seconds of boot time is lost while scanning for USB devices. Such a long delay is very annoying, but this is necessary for USB stick recovery. I wish there was another way...

btw. I was able to eliminate most of this by disabling the OHCI controller, see commit https://github.com/Openvario/meta-openvario/pull/267/commits/9b3974a4a4ac40747f68e67bc059098f3cdf5a10 in this PR

We're down to 14 seconds from power on to ovmenu.

MaxKellermann:
```
# systemd-analyze      
Startup finished in 1.275s (kernel) + 5.785s (userspace) = 7.061s 
multi-user.target reached after 5.688s in userspace
```

MaxKellermann:
Second measurement was 12 seconds.
My goal is less than 10 seconds from power on to XCSoar fully started and audio vario running.

August2111:
For most users who have the OpenVario installed in their glider, the fastest possible boot time is a very important criterion, so everything eas Max does in this matter is to be supported!

linuxianer99 (approved):
Looks good to me !
Thanks Max for all this effort !!

kedder (approved):
LGTM

MaxKellermann:
> Second measurement was 12 seconds. My goal is less than 10 seconds from power on to XCSoar fully started and audio vario running.

Latest measurement: 9.5 seconds from power on (cold boot) to OpenVarioMenu. More PRs to come.

mihu-ov:
Below 10s is almost like "power on" instead of "booting", And if I understand correctly all of this without detrimental (rather the opposite) impact on future maintainability. That´s great!

MaxKellermann:
> Below 10s is almost like "power on" instead of "booting"

Right, and next step: get rid of psplash and its progress bar. What's the point of a progress bar if it boots almost instantly? Loading this second splash screen only delays the boot process.

> And if I understand correctly all of this without detrimental (rather the opposite) impact on future maintainability. That´s great!

Yes.

PR #269, opened 2022-02-11
Linux kernel 5.15.22

MaxKellermann:
Updates to the latest Linux 5.10.y kernel. Next time, I'll update to the latest LTS kernel 5.15.y.

By disabling SATA support, this saves 200ms kernel boot time. Does anybody use SATA on an OpenVario?

New record: power on to ovmenu in 11.3 seconds.


MaxKellermann:
> Next time, I'll update to the latest LTS kernel 5.15.y.

This is ... "next time". Updated to 5.15.22 in a rather trivial commit.
The massive cleanup that precedes it was the difficult part ...

linuxianer99:
> Updates to the latest Linux 5.10.y kernel. Next time, I'll update to the latest LTS kernel 5.15.y.
> 
> By disabling SATA support, this saves 200ms kernel boot time. Does anybody use SATA on an OpenVario?
> 
> New record: power on to ovmenu in 11.3 seconds.

Think SATA is never used.

Great boot time !! Congratulations !!

linuxianer99 (commented):
Looks good !
Just one question: Why are we going back to a "own" kernel recipe instead of use the bbappend ??

MaxKellermann:
> Why are we going back to a "own" kernel recipe instead of use the bbappend ??

The sunxi recipe is stuck with 5.10.2, and there's no perspective this will change. Also the sunxi recipe is full of useless stuff.
Then there's the linux-yocto recipe, also full of useless stuff.
And now, look at how simple the new recipe is. It inherits from "kernel.bbclass" and "kernel-yocto.bbclass", and contains just what is special for OpenVario. A bbappend would not be simpler. By inheriting from those bbclasses, we are better than "bbappending".

linuxianer99 (approved):

linuxianer99:
> > Why are we going back to a "own" kernel recipe instead of use the bbappend ??
> 
> The sunxi recipe is stuck with 5.10.2, and there's no perspective this will change. Also the sunxi recipe is full of useless stuff. Then there's the linux-yocto recipe, also full of useless stuff. And now, look at how simple the new recipe is. It inherits from "kernel.bbclass" and "kernel-yocto.bbclass", and contains just what is special for OpenVario. A bbappend would not be simpler. By inheriting from those bbclasses, we are better than "bbappending".

got it! Thanks for explaination

MaxKellermann:
Linux 5.15.23 was just released - I amended the patch.

MaxKellermann:
Amended one patch to remove an obsolete variable, and added another patch to remove `DEFAULTTUNE`. Not exactly related to the kernel, but doesn't belong anywhere else ...

meta-ov/recipes-kernel/linux/files/openvario.cfg
```diff
@@ -25,7 +25,6 @@ CONFIG_UEVENT_HELPER=y
 CONFIG_INPUT_EVDEV=y
 # CONFIG_INPUT_FF_MEMLESS is not set
 # CONFIG_INPUT_MOUSE is not set
-CONFIG_INPUT_UINPUT=y
```
kedder, line 30: I think touchscreen handling is based on uinput (`ts_uinput` daemon). If it is the same uinput, suggest to leave it enabled.

kedder (changes requested):

meta-ov/recipes-kernel/linux/files/openvario.cfg
MaxKellermann: It is still enabled; the line was duplicate, and this commit just removes one copy.

kedder (approved):

PR #270, opened 2022-02-13
Timeout usbstick automount after 2 seconds

kedder:
When user tries to perform any operation that involves USB stick (e.g.
"Download IGC") and USB stick is not inserted, openvario appears to
"freeze" - becomes unresponsive, with no way to recover.

This is because it waits for /dev/sda1 device to appear in order to
mount it.  With the default 90s timeout the system appears to be stuck.

This lowers the timeout to 2 seconds to "unstuck" the system quickly
enough before user gets frustrated.

### DIFF
--- a/meta-ov/recipes-support/autofs-config/files/usb-usbstick.mount
+++ b/meta-ov/recipes-support/autofs-config/files/usb-usbstick.mount
@@ -1,5 +1,6 @@
 [Unit]
 Description=Automatically mount connected USB drives
+JobTimeoutSec=2
 
 [Mount]
 What=/dev/sda1


MaxKellermann:
`JobTimeoutSec` is the generic timeout for a start job, but mount units have a `TimeoutSec` setting which seems more appropriate. I don't know the exact semantics of both, I just guess "more specific" = "better".
I wonder if we can avoid waiting for a device completely, i.e. never wait, and fail immediately if the device doesn't exist. Because ... what's the point of waiting?

And actually, I wouldn't use the auto-mounter at all. I changed to systemd automount not because I think auto-mounting is fine, but only because I think systemd can do that feature better than the old "autofs" daemon. I think it's better to explicitly mount in all programs that wish to access the USB stick, and more importantly: explicitly unmount when finished.

kedder:
`TimeoutSec` is a timeout for `mount` command, but in this case `mount`command never executes: `systemd` actually waits for `/dev/sda1` device to appear. And it will wait for 90s before failing because it will never appear. IOW, `TimeoutSec` doesn't work for this purpose.

Automounter is nice in a way because it handles all the edge cases.  E.g. it makes sure device is correctly unmounted even if app that reads the device crashes because the stick was taken out in the middle of copy operation.

kedder:
> I wonder if we can avoid waiting for a device completely, i.e. never wait, and fail immediately if the device doesn't exist. Because ... what's the point of waiting?

That's what I do in ovshell - wait for device to appear before looking into the `/usb/usbstick` directory. Probably something like this can be done in the ng-menu script as well. But at least things are not get IO blocked if something (e.g. `opkg update`) tries to read the mount point directly.

MaxKellermann:
> `TimeoutSec` is a timeout for `mount` command, but in this case `mount`command never executes: `systemd` actually waits for `/dev/sda1` device to appear. And it will wait for 90s before failing because it will never appear. IOW, `TimeoutSec` doesn't work for this purpose.

You're right. Though, reading the systemd source code, it looks like "JobRunningTimeoutSec" is better. The fstab generator translates  "x-systemd.device-timeout" ("Configure how long systemd should wait for a device to show up before giving up") to "JobRunningTimeoutSec": https://github.com/systemd/systemd/blob/main/src/shared/generator.c#L281

linuxianer99 (approved):
OK for me.
But we should really think about the solution Max mentioned: Only mount the stick if necessary (an app needs it)

kedder:
@MaxKellermann `JobRunningTimeoutSec` doesn't work for the same reason as `TimeoutSec` - the job isn't actually running - it is waiting for dependent job (the device) to finish.

kedder:
Can we merge this?

kedder:
Assuming no objections.

linuxianer99:
@MaxKellermann ??

MaxKellermann:
> @MaxKellermann `JobRunningTimeoutSec` doesn't work for the same reason as `TimeoutSec` - the job isn't actually running - it is waiting for dependent job (the device) to finish.

Does this mean the systemd generator is buggy?

MaxKellermann:
> I don't see why you would imply that.

Because:

> The fstab generator translates "x-systemd.device-timeout" ("Configure how long systemd should wait for a device to show up before giving up") to "JobRunningTimeoutSec": https://github.com/systemd/systemd/blob/main/src/shared/generator.c#L281

But I found the answer already: this sets the device's timeout, not the mount's timeout.

PR #271, opened 2022-02-15
openvario-logo: fix resolution for 7-inch texim displays (CH070)

Scumi:


mihu-ov:
Confirm, 800x480 is the correct resolution for the Texim CH070 version.

MaxKellermann (approved):

PR #284, opened 2022-02-28
sensord: use systemd socket activation

MaxKellermann:
Related to https://github.com/Openvario/sensord/issues/42 and https://github.com/Openvario/sensord/pull/43


linuxianer99 (approved):
lgtm

PR #290, opened 2022-04-05
Update local.conf

Blaubart:
Updating local.conf to include i2cdetect and chattr to the image again

### DIFF
--- a/conf/local.conf
+++ b/conf/local.conf
@@ -241,3 +241,6 @@ PACKAGECONFIG:append:pn-nativesdk-qemu = " sdl"
 # track the version of this file when it was generated. This can safely be ignored if
 # this doesn't mean anything to you.
 CONF_VERSION = "2"
+
+IMAGE_INSTALL:append = "i2c-tools"
+IMAGE_INSTALL:append = "e2fsprogs"


Blaubart:
Hmmm, I don't know why the check dosen't work:
NOTE: Runtime target 'i2c-toolse2fsprogs' is unbuildable, removing...

I testest it yesterday, and it works

August2111:
There is a missing blank character before 'i2c-tools' and 'e2fsprogs' I think!
But I don't know, if this is really wanted for all OpenVario images, f.e. at devices without sensorboard?

Blaubart:
> But I don't know, if this is really wanted for all OpenVario images, f.e. at devices without sensorboard?
i2cdetect is really important if you build your OV by your self! So why shouldn't we include it generally?? If some users dosen't need it, it dosen't matter.

linuxianer99:
Please move the apps to be installed in the image to the image recipe.
openvario-base-image.bb

Local.conf will be generated at the init of the build system ... so no good place ;-)

Can you please check, what dependencies are caused by this two tools ?
Max spent a lot of effort to reduce the image size, so i don't want to blow it up ...

Blaubart:
> Can you please check, what dependencies are caused by this two tools ?

Sorry, I don't know how to check. At Ubuntu I can use "apt show e2fsprogs" :
Pre-Depends: libblkid1 (>= 2.17.2), libc6 (>= 2.14), libcom-err2 (>= 1.42~WIP-2011-10-05-1), libext2fs2 (= 1.44.1-1ubuntu1.3), libss2 (>= 1.34-1), libuuid1 (>= 2.16)

i2cdetect dosen't exist in Ubuntu

linuxianer99:
https://docs.yoctoproject.org/bitbake/bitbake-user-manual/bitbake-user-manual-intro.html#generating-dependency-graphs

Blaubart:
I hope these files are helping:
https://vdrblaubart.dynvpn.de/s/fn6gnT4gF6GeTbL

linuxianer99:
so i2c-tools adds:
db, gdbm, perl

Do we really want to add perl again ??

Blaubart:
I can not judge over this. But, like many others, I build all my OpenVarios myself. Anyone who has ever soldered a sensor board knows how difficult some sensors are to solder. Without the tool i2cdetect you don't have any reasonable feedback as to which sensors work and which don't. So no clear diagnosis is possible. So for everyone who builds themselves, the tool is indispensable!

mihu-ov:
I agree i2cdetect is important for debugging sensor issues but this doesn´t require adding it to the image. One of the old images with i2c tools can be used for debugging, or it can be built as a package and installed if required.

linuxianer99:
> I agree i2cdetect is important for debugging sensor issues but this doesn´t require adding it to the image. One of the old images with i2c tools can be used for debugging, or it can be built as a package and installed if required.

I agree with @mihu-ov . As discussed here ([Link](https://github.com/Openvario/meta-openvario/discussions/181), we should think, which tool should be integrated in which variant ...

Blaubart:
OK, I understand your concerns. But please also consider that not everyone who builds their own OV has the same level of knowledge in the field of software as you do! I've been dealing with the OV for quite a while, but I have no idea how to post-install i2cdetect as a package. There should also be clear instructions for this, e.g. at https://www.openvario.org/doku.php?id=projects:series_00:electrical_tests:sensorboard

August2111:
> There should also be clear instructions for this, e.g. at https://www.openvario.org/doku.php?id=projects:series_00:electrical_tests:sensorboard

Ok, this is my part... There are a lot of things to do on the OpenVario-Homepage, and I don't like to do things twice. But now it is a good state reached to do the rework - but I can only do this step by step...
Like Michael and Timo I prefer a clean basic image - with maybe  to install some additional things as an extension, and not a blowing up of the basic image for an 'normal' user!

MaxKellermann:
The first two commits in this PR are bad, and the third one contains fixups for the first two. The PR should not be merged that way.

I don't know about i2c-tools, but I'd rather not include e2fsprogs just for chattr. I think the chattr trick is a bad kludge that should not be promoted or supported.

Blaubart:
> ... or it can be built as a package and installed if required.

Can you explain to me how to build and install the additional packages?

PR #308, opened 2022-05-23
Update to the latest kirkstone

kedder:
This fixes the build from scratch, that is failing on libgcc otherwise.


linuxianer99:
Did the build really fail ??
I did a testing build yesterday working well ...

kedder:
Dunno, 2 packages failed to build for me with weird compilation and packaging errors (exact error messages are lost ATM). Upgrading fixed it. May be an accident, but keeping things updated should not hurt anyways.

PR #315, opened 2022-06-25
improve backup and restore procedure

Blaubart:
This backup script stores all XCSoar settings and relevant OpenVario settings like:
- brightness of the display
- rotation
- Touch screen calibration
- language settings
- dropbear settings

backups are stored at USB stick at:
openvario/backup/"MAC address of eth0"/

So you can store more than one backup on the same stick!


7lima:
@Blaubart
Do you think we should include your merge request comments from above into the scripts' header comments as an explanation of the features? - This wouldn't make them any slower ;-)

Blaubart:
> Do you think we should include your merge request comments from above into the scripts' header comments as an explanation of the features? - This wouldn't make them any slower ;-)

Good idea!!

lordfolken:
In general i recommend https://shellcheck.net/ as a sanity check.

7lima:
@lordfolken 
Thank you a lot for your tip. The checker showed me some non-obvious defects which I've now fixed.
I think I'll use shellcheck from now on regularly although I'm programming shell scripts for 38 years now.

linuxianer99:
So i am ok with the content, but not with the 40 commits for this content ... you want just add 2 files and modify an existing.

Please restructure the PR.

7lima:
> 

OK, I'll do.
So let's play squash!

MaxKellermann:
This is still a huge mess, and it's getting bigger, not smaller.

Blaubart:
Yes, that's because we found some bugs and start improving the code a bit. I think today it is finished and we can do squashing

MaxKellermann:
If you ever need to squash, then your workflow is flawed.

PR #317, opened 2022-07-11
improve backup and restore procedure

Blaubart:
This backup script stores all XCSoar settings and relevant OpenVario settings like:

- brightness of the display
- rotation
- touch screen calibration
- language settings
- dropbear settings

backups are stored at USB stick at:
- openvario/backup/"MAC address of eth0"/

So you can store more than one backup on the same stick!

We add system info script, too

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,10 @@
 OpenVario*rootfs.img.gz
 .DS_Store
 bitbake.lock
+.project
 cache/
 downloads/
 sstate-cache/
 tmp/
 build/
+.settings/
--- a/meta-ov/recipes-apps/ovmenu-ng-skripts/files/system-info.sh
+++ b/meta-ov/recipes-apps/ovmenu-ng-skripts/files/system-info.sh
@@ -1,0 +1,92 @@
+#!/bin/sh
+#
+# system-info.sh
+# System info script gives an overview which packages and which versions are instaled
+#
+# Created by Blaubart,        2022-02-08
+#
+
+# collect info of system and more installed packages
+IMAGE_VERSION=$(cat /etc/os-release | grep VERSION_ID | cut -d '=' -f 2)
+KERNEL_VERSION=$(uname -r)
+XCSOAR_MAPS_VERSION=$(opkg list-installed xcsoar-maps* | cut -d '-' -f 4)
+XCSOAR_MENU=$(opkg list-installed xcsoar-menu* | cut -d '-' -f 3)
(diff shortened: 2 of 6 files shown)


MaxKellermann:
This is now a different kind of horrible mess.

Blaubart:
Why??

MaxKellermann:
The diff is also very confusing. The commit does many different and independent things, but one commit should only do one thing.

MaxKellermann:
What this PR seems to be doing:
- add stuff to .gitignore
- add some system-info feature
- new script transfer-xcsoar
- removing several old scripts
- lots of whitespace changes in ovmenu-ng.sh
- moving code around
- at the same time, modify the code that was moved around
- and maybe more, I can't see, I'm so confused

All of this is very confusing, and code review is nearly impossible.

7lima:
So we have 2 elements here, the backup/restore and the system info which have both been completely reworked. Furthermore the file and function names have been renamed. I understand that it's a lot to swallow as a single chunk. Would you prefer if we divided the PR into 2 or 3 parts?

MaxKellermann:
Note that I don't manage this project, I just state my opinion here (without being asked to).
On my projects, I expect contributors to write PRs where each commit is self-contained and documented, each commit is known-good (no known defects, and no fixup commits in the same PR), and each commit contains only one change (which may be spread across multiple files, so all code modifications belonging to that change are in the one commit, but not more than one unrelated change in one commit).

If you were about to improve the backup procedure, that should be one commit, but that commit should not rename scripts, because that would obfuscate the changes and would make reviewers unable to see what you were trying to improve.
If you were about to add a system info feature, that doesn't belong in a commit that also improves backup/restore.
You could suggest adding patterns to .gitignore, but that should be a separate commit, and the commit message should explain why this pattern is needed. (If this pattern is specific to your own workflow, it should be in your own private .gitignore file and not merged in this public repository.)

Many different changes can be put in one PR, but as separate commits.

Blaubart:
Hi Max,

thanks for your explanations!! We are working on taking this into account in the future and making it better.

7lima:
I'll separate the different functional items into coherent commits building upon each other. Then you may have a look again.

linuxianer99:
Hi All,
i share the opinion of @MaxKellermann . 
So have a different features in different commits also makes a potential bug fixing a lot easier, if this comes necessary ...

Having multiple commits in on PR is fine ...

You can also have a look here:
https://github.com/Openvario/meta-openvario/wiki/Code-Contribution-Rules

PR #318, opened 2022-07-19
Openvario Freeze fix

tb59427:
This pull requests fixes random Openvario freezes (hopefully all of them) that have happened seemingly out of the blue (Openvario issue #304) . The root cause for these freezes was that CPU voltage for the lowest 2 CPU speeds (144 and 312 Mhz respectively) is originally set at 1.0V in sun7i-a20.dtsi. Apparently this voltage is too low and  leads to hardware instability on the cubieboard - at least in the use case for Openvario. 

Inital analysis lead to fixing CPU speed (thus effectively disabling adaptive CPU speed done by the cpu_freq driver). It wasn't the CPU speed fixing, though, but the corresponding fixing of CPU voltage to 1.1V up to 1.4 at the top speed. That actually erased the instability. The increased power consumption of the CPU at 1.1V is neglectable (also according to [this analysis](https://linux-sunxi.org/Cpufreq) - see section "Practical impact on temperature/consumption"). Once that analysis was concluded the fix is (as so often) fairly simple:

The fix itself consists of two files
1. a patchfile "0004-1.1V-fix-for-sun7i-a20.dtsi.patch" that changes the minimum CPU voltage to 1.1V in "arch/arm/boot/dts/sun7i-a20.dtsi" of the linux source tree and
2. the inclusion of that patchfile in the kernel build recipe

BEWARE: this patchfile (as the other 3 patchfiles currently used in the build as well) needs careful checking with every new kernel that is being used for the Openvario build as the underlying file "arch/arm/boot/dts/sun7i-a20.dtsi" could change in future kernels. Also: should Openvario's hardware switch over to other arm boards, this fix may either be obsolete or in need of adaptation.

NB: I have started this pull request as a draft pull request in order to
a) give people more experienced than me the opportunity to review and comment and
b) allow for more community testing of the fix prior to actually merging it into the code base and
c) because I am a complete github rookie and didn't want to create chaos :-)


mihu-ov:
Thank you for tracing and tackling this issue! To me, your analysis makes a lot of sense and I think we should implement this in our OV image right now and see if it fixes the freezes for everyone. We really need the wider user base to be sure. The only possible downside I see is power consumption, and stability over slightly increased power consumption is a clear choice in my opinion. We could even think about increasing minimum frequency to 312MHz.

linuxianer99 (approved):

MaxKellermann:
The two commits should be one commit, and the new patch file has a bad description.

tb59427:
1. how do I fold these two commits into one? (I am a git rookie as stated earlier)?
2. what constitutes a good description for a patch file?

MaxKellermann:
1. there are many ways to do that, but I recommend using stgit (https://stacked-git.github.io/guides/usage-example/ https://stacked-git.github.io/guides/tutorial/). stgit is one of the tools I cannot imagine ever working without.
2. you could copy your commit message as the patch description

I just noticed that your commit message is also very brief with no real explanation, while this PR's text is very long. Writing PR texts is a complete waste of time, because they get lost, they are only in GitHub's proprietary database. Only commit messages remain forever.

Having good documentation (in commit messages and/or code comments) is utterly important, even more so if it's about an obscure hardware problem. Pieces of code which work around such issues need to be documented, or you leave other people in a few years wondering what the hell this is about, and that kind of working will turn every software project into a huge pile of workarounds stacked on top of countless other workarounds, and eventually nobody understands anything anymore. (This is how the XCSoar code base looked like when I joined the project. That was an awful experience.)

Bottom line is: if figuring out a problem took you a long time, please take some more time explaining your insights - it's important. The more, the better.

tb59427:
obviously I am too stupid to use git. I now have added more description to the commits, but the system has generated new commits despite [me following this guide](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message). Git remains a mystery to me :-(

tb59427:
It's only getting worse with me and git. Can someone please confirm the approach:
1. fork this OV repository
2. do the changes in the forked repo (do I have to have a new branch for this in my local repo?)
3. commit the changes (trying to do all changes in one commit  - that works only from the CLI right? Web UI always seems to want to do separate commits)
4. pull request.

I probably cancel this pull request and do it all over again (luckily not too much of a change anyway).

linuxianer99:
> It's only getting worse with me and git. Can someone please confirm the approach:
> 
> 1. fork this OV repository => yes
> 2. do the changes in the forked repo (do I have to have a new branch for this in my local repo?) => Yes, you need a branch
> 3. commit the changes (trying to do all changes in one commit  - that works only from the CLI right? Web UI always seems to want to do separate commits) => I only use the cli (git add, git commit, git push, ...)
> 4. pull request. => yes
> 
> I probably cancel this pull request and do it all over again (luckily not too much of a change anyway).

Maybe you can have a look on these:
https://opensource.com/article/19/7/create-pull-request-github
It is the first google hit ..

tb59427:
I'll redo this pull from a fresh fork and will try to incorporate your suggestions.

PR #322, opened 2022-07-25
Change of CPU OPP voltages to eliminate freezes

tb59427:
This change, when applied, changes the OPP voltages for both 144MHz and 312MHz from 1V to 1.1V when creating Openvario images. This fixes freezes previously encountered.

Reason for this change

Starting some time in 2021 Openvario users experienced inexplicable sudden crashes of their systems. I started analysing these issues by asking as many users as possible to document their hardware/software setup (see here https://docs.google.com/spreadsheets/d/1GQ6Q5XrnuuqNc-38CmHEat-4Oquq_l_X-1m1NVBfuBU/edit?usp=sharing) in order to find some sort of systematic behind these crashed (I failed :-). The results were inconclusive. Next I set up an OpenVarion test system with console output and tested various linux versions each with kernel debugging enabled. This lead to identification of the cpu_freq driver as being a possible culprit for these freezes. Many crashes happened as a consequence of cpu_freq changing CPU speed. A first workaround was tried fixing CPU speed at a defined frequency to stop cpu_freq from switching CPU speeds (for details of this work around see here https://github.com/tb59427/openvario-cpu-freq-fix). this work around significantly improved stability at the cost of a fixed cpu speed and potentially high power usage.

Further analysis revealed that the kernel tried to access parts of a struct with invalid addresses while access to the very same struct succeeded a few instructions earlier. With many insights from Samuel Holland of the linux-sunxi mailing list, we concluded that it wasn't the frequency but most likely OPP voltage settings of the CPU that created hardware instability resulting destroyed or wrong addresses. See thread here: https://groups.google.com/g/linux-sunxi/c/iAcd0PxOuOs .

Per the A20 SDK (https://linux-sunxi.org/A20) (see section DVFS) the DVFS OPP voltages for 144000 should be 1050000mV and for 312000 it should be 1100000mV. These voltages, however, are set to be 1000000mV for each of the affected frequencies in the upstream linux kernel sources. These low voltages obviously lead to hardware instabilities observed.

There was a commit for the linux source tree already in 2015 (a15b80f) which changed OPP voltages according to the A20 SDK documentation - however only specifically for the bananapi variant of the A20 board (arch/arm/boot/dts/sun7i-a20-bananapi.dts). For some unknown reason this patch never made it to the core sun7i-a20.dtsi file which is imported by all sun7i A20 dts files (but can be overwritten by dts files specifically designed for an A20 board make.

The suggestion on the linux-sunxi mailinglist was to increase OPP voltage by 100mV (thus pushing the OPP voltage for the lowest frequency even 50mV higher than the A20 SDK recommendation).

Weighing A20 SDK information against Samuel's sentiment (who has dealt with many A20 boards) I decided to go for the 110000mV for the lowest two frequencies. As the A20 SoC has two cpus OPP voltages have been increased for both frequencies.

What this change does

This change changes OPP voltages from 1V to 1.1V for the lowermost two CPU frequencies both for CPU0 and CPU1. These OPP voltages are set in arch/arm/boot/dts/sun7i-a20.dtsi of the linux source tree. This file gets imported by all other A20 dts files for specific variants of the A20 SoC. The OpenVario kernel recipes also include this file (via a chain of inclusions starting with "meta-openvario/meta-ov/recipes-krenel/linux/files/openvario-common.dts"). The change is applied during bitbake of the kernel by applying the patchfile to arch/arm/boot/dts/sun7i-a20.dtsi prior to compilation. Additionally the kernel bitbake file (meta-openvario/meta-ov/recipes-kernel/linux/linux-openvario_5.17.5.bb gets changed to apply the patchfile.

Other considerations

In theory I could have also changed the openvario-common.dts file for the OpenVario build in meta-ov/recipes-kernel/linux/files and overwritten the definitions imported from arch/arm/boot/dts/sun7i-a20.dtsi of the linux source tree. This would have avoided patching a kernel file during the bitbake process. On the other hand we would have buried a kernel dependency by copying a fraction of the kernel code into the OpenVario build. Should the kernel substantially change in the future - in particular arch/arm/boot/dts/sun7i-a20.dtsi we may run the risk over not noticing and running into inexplicable errors. Isolating this patch into a patch to a kernel file indicates that there is a dependency and with every new kernel building the basis for OpenVario we are now forced to review arch/arm/boot/dts/sun7i-a20.dtsi and whether the patch is still correct. I preferred this approach over "overwrite" in
openvario-common.dts.

Beware

This change needs revisiting everytime the kernel version for Openvario get's changed as new / different kernel versions may contain a different dtsi file for the A20.

It remains to be seen how well this change eliminates crashes. My test system runs rock stable with this change, other users on the xcsoar forum report this, too. But there are also hints that 1.1V may not be enough. It could well be that further increase (by 100mV) of the OPP voltages may be needed. Using the xcsoar forum to collect feedback will hopefully provide more clarity about this question.

### DIFF
--- a/meta-ov/conf/machine/openvario-43-rgb.conf
+++ b/meta-ov/conf/machine/openvario-43-rgb.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-43-rgb.dts"
+KERNEL_DEVICETREE = "openvario-43-rgb.dtb"
 
 SHORT_OV_MACHINE= "AM43"
--- a/meta-ov/conf/machine/openvario-57-lvds-DS2.conf
+++ b/meta-ov/conf/machine/openvario-57-lvds-DS2.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-57-lvds-DS2.dts"
+KERNEL_DEVICETREE = "openvario-57-lvds-DS2.dtb"
 
 SHORT_OV_MACHINE= "CH57_2"
--- a/meta-ov/conf/machine/openvario-57-lvds.conf
+++ b/meta-ov/conf/machine/openvario-57-lvds.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-57-lvds.dts"
+KERNEL_DEVICETREE = "openvario-57-lvds.dtb"
 
 SHORT_OV_MACHINE= "CH57"
--- a/meta-ov/conf/machine/openvario-7-AM070-DS2.conf
+++ b/meta-ov/conf/machine/openvario-7-AM070-DS2.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-7-AM070-DS2.dts"
+KERNEL_DEVICETREE = "openvario-7-AM070-DS2.dtb"
 
 SHORT_OV_MACHINE= "AM70_2"
--- a/meta-ov/conf/machine/openvario-7-CH070-DS2.conf
+++ b/meta-ov/conf/machine/openvario-7-CH070-DS2.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-7-CH070-DS2.dts"
+KERNEL_DEVICETREE = "openvario-7-CH070-DS2.dtb"
 
 SHORT_OV_MACHINE= "CH70_2"
--- a/meta-ov/conf/machine/openvario-7-CH070.conf
+++ b/meta-ov/conf/machine/openvario-7-CH070.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-7-CH070.dts"
+KERNEL_DEVICETREE = "openvario-7-CH070.dtb"
 
 SHORT_OV_MACHINE= "CH70"
--- a/meta-ov/conf/machine/openvario-7-PQ070.conf
+++ b/meta-ov/conf/machine/openvario-7-PQ070.conf
@@ -4,6 +4,6 @@
 
 require ov-cubieboard2.inc
 
-KERNEL_DEVICETREE_SOURCE = "openvario-7-PQ070.dts"
+KERNEL_DEVICETREE = "openvario-7-PQ070.dtb"
 
 SHORT_OV_MACHINE= "PQ70"
--- a/meta-ov/conf/machine/ov-cubieboard2.inc
+++ b/meta-ov/conf/machine/ov-cubieboard2.inc
@@ -11,5 +11,3 @@ MACHINE_FEATURES:remove = "apm"
 
 IMAGE_CLASSES += "sdcard_image-openvario"
 IMAGE_FSTYPES = "openvario-sdimg"
-
-KERNEL_DEVICETREE = "openvario.dtb"
--- a/meta-ov/recipes-bsp/u-boot/files/bootenv.ini
+++ b/meta-ov/recipes-bsp/u-boot/files/bootenv.ini
@@ -18,12 +18,6 @@ done;
 [bootcmd_mmc0]
 echo Try to boot from mmc ...;
 run set_bootargs;
-
-# hard-coded DeviceTree file name because u-boot uses the standard
-# Cubieboard 2 DeviceTree, but the Linux kernel needs a custom
-# DeviceTree
-setenv fdtfile openvario.dtb;
-
 load mmc 0:1 ${fdt_addr_r} ${fdtfile};
 load mmc 0:1 ${kernel_addr_r} uImage;
 bootm ${kernel_addr_r} - ${fdt_addr_r}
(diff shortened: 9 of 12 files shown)


tb59427:
So guys, this should be better now...pls check

linuxianer99 (approved):
Fine for me

tb59427:
And again git has surprised me with additional commits. I have watched endless videos and tutorials but seem to be unable to change or edit an existing commit - instead I am generating more commits all the time. Despite following @MaxKellermann's suggestion to work with stg. I am stuck. Somebody who better understands and masters git needs to merge my patch into OV. I am incapable of mastering  git apparently.

linuxianer99:
@tb59427 : What was you intention with the latest commit ??

Each time, you push to your private fork, the PR is based on, the commit will be added.
changing a commit is call "amend" in git .. but this is just for the commit message as far as i know ...

tb59427:
@MaxKellermann wanted a more descriptive title of the commit message. I changed that (at least now visible in my local copy). Why git has created a shitload of commits out of that change is beyond my comprehension. I used stgit to edit the commit message.

As far as I understand "amend" only works as long as the commit hasn't been pushed.

Anyway in terms of the most important piece, the patch, nothing has changed. It's all about the commit messages. Why git isn't allowing editing of commit messages is also beyond my comprehension....

linuxianer99:
> @MaxKellermann wanted a more descriptive title of the commit message. I changed that (at least now visible in my local copy). Why git has created a shitload of commits out of that change is beyond my comprehension. I used stgit to edit the commit message.
> 
> As far as I understand "amend" only works as long as the commit hasn't been pushed.
> 
> Anyway in terms of the most important piece, the patch, nothing has changed. It's all about the commit messages. Why git isn't allowing editing of commit messages is also beyond my comprehension....

You can change with amend in your private local repository and do a force-push to your private remote repo.
So a commit is a commit. Change the comment afterwards, is not allowed ;-)

tb59427:
yeah I figured that much :-P

linuxianer99:
For example:

https://levelup.gitconnected.com/how-to-rewrite-the-latest-commit-with-git-amend-48a8d5f27758

tb59427:
Yeah I have read that. The question is - can I remove the unnessary two commits and just change the title / subject of the original commit? If that's possible let me know. Else I am stuck and some git guru needs to fix the mess

linuxianer99:
Have a look here:

https://ncona.com/2011/07/how-to-delete-a-commit-in-git-local-and-remote/

tb59427:
That's the approach I tried earlier and that created the inital mess...I am using google, too. It just seems that I don't seem to be able to generate the result forseen in those tutorials....

linuxianer99:
> That's the approach I tried earlier and that created the inital mess...I am using google, too. It just seems that I don't seem to be able to generate the result forseen in those tutorials....

Maybe you can create a Dummy repo .. just create some commits and learn, how this works ? 
I also learned git by just playing arround ...

tb59427:
Yeah, I need to do that :-) My original focus was to fix OV freezes, though :-P The git issues started to pop up as a side effect, really - and it feels I have now spent more time struggling with git than with the f...ing kernel bug :-P

tb59427:
> Have a look here:
> 
> https://ncona.com/2011/07/how-to-delete-a-commit-in-git-local-and-remote/
> 
> so i am also no git-guru ... just using google ;-)

Re-did this again - now my local repo looks right. `git push origin +master` shows Everything up-to-date - but it looks like nothing has changed in my remote repo nor the OV repo...

linuxianer99:
Think you have to push your feature branch (OPP-1.1V-voltage-fix) not master.

tb59427:
Yeah I did this. Now git wants to push 4 more commits and pull 7. I'm giving up. Nothing has changed and git keeps pushing and pulling like crazy....

DanD222:
Maybe try a Git GUI client... something like Git Kraken perhaps? Worth giving a shot and seeing if that helps.

tb59427:
> Maybe try a Git GUI client... something like Git Kraken perhaps? Worth giving a shot and seeing if that helps.

That only shows me graphically that it wants to pull 7 changes from the repo and push 4 - all of which are already in the repo.

DanD222:
OK.. what I find it helpful for is getting my local repo commits into shape using whatever tools are necessary (rebase, squash, amend), and once I see that it looks good in the GUI I can do a force-push to overwrite what's in my online repo, which will also update the pull request.

tb59427:
@DanD222 - I'd love to  do that. Only everytime I try, it's getting worse...

linuxianer99:
> OK.. what I find it helpful for is getting my local repo commits into shape using whatever tools are necessary (rebase, squash, amend), and once I see that it looks good in the GUI I can do a force-push to overwrite what's in my online repo, which will also update the pull request.

Yes !! The correct way to update a PR !

tb59427:
So this is what it looks like locally (`git log --pretty=oneline --abbrev-commit`):
```
640b93b (HEAD -> OPP-1.1V-voltage-fix) Change CPU OPP voltages from 1V to 1.1V for lowest two CPU speeds
cc1e80c linux-openvario: keep the original DT filename
4b2e26e linux-openvario: remove useless kernel-devicetree dependency
7d40e85 openvario-base-image: broader glob for deleting the redundant kernel image
42d5dea Merge pull request #308 from kedder/master
f7ce3a6 Update to the latest kirkstone
```

To me this looks right. Yet `git status` says this
```
On branch OPP-1.1V-voltage-fix
Your branch and 'origin/OPP-1.1V-voltage-fix' have diverged,
and have 4 and 7 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

nothing to commit, working tree clean
```

I find this completely confusing and have no idea how to sort this and when I pull/push I will probably just generate more useless commits.

DanD222:
What's I also find useful is if you get stuck in a mess, you can reset your offline repo to the master branch (called "upstream", i.e. https://github.com/Openvario/meta-openvario), and start again. Then only once you're happy with the commits do you need to push to your online repo (https://github.com/tb59427/meta-openvario). This way you don't need to close a PR and start a new one; force-pushing will overwrite what you've already done in the online version of your branch (called "origin").

tb59427:
Now this stupid git system show commits from Max as my commits....I'll cancel this pull request (again). It's only getting worse with me and git. Somebody who feels comfortable using git can have the files from me and do the pull request.

linuxianer99:
@tb59427 You really give up ??!! ;-)

So easiest way:
- Create a new branch in your private fork of meta-openvairo (name it different than the existing)
  - git checkout -b <some branch>
- change the files ...
- Do git add ... of the file
- Do git commit and enter your commit message you want
- Do git push
  - git push –set-upstream origin <your new branch name>

So ... come on ... give it a try ;-)

PR #325, opened 2022-08-11
Update XCSoar, kernel, Yocto

MaxKellermann:


mihu-ov:
Just a general question: should we use the latest stable linux kernel, or the latest longterm supported kernel for best stability?

linuxianer99 (approved):
LGTM

MaxKellermann:
> Just a general question: should we use the latest stable linux kernel, or the latest longterm supported kernel for best stability?

That's a question of taste.
If you only use longterm kernels, you have bigger steps between the versions, which makes upgrades much more complicated, and you keep postponing the updates because they are so complicated.
Using latest stable means smaller steps, but more often. That quickly becomes an easy routine. I prefer that.

mihu-ov:
Thank you for the explanation!

August2111:
From my point of view it should be merged asap... (XCSoar 7.24!)

PR #328, opened 2022-09-06
initial Raspberry Pi support (experimental)

MaxKellermann:


bomilkar:
How do you envision to implement the different display parameters? (size, pixel clock, sync, ...)
It would be nice if this can be accomplished by means of a config file (on a FAT32 partition) like /boot/config.txt in Raspbian. That way we wouldn't need display specific images.
Correct me if I'm wrong, one of the primary goals is to operate the "old" displays on a Raspberry Pi.

PR #330, opened 2022-09-19
Further increase minimum CPU voltage to 1.2V

tb59427:
This change, when applied, changes the OPP voltages for both 144MHz and 312MHz from 1V to 1.2V when creating Openvario images. This fixes freezes previously encountered. It is a "follow up" change to PR #324. It turned out that even at 1.1V OV occasionally crashed.

For more information on the why and how of this change please see comment in PR #324


MaxKellermann:
I think commit messages should refer to commit ids, not to PR numbers, because those are proprietary GitHub things out of git's scope

mihu-ov:
I think further increasing voltage for the low frequencies is good but avoiding them by additionally setting a minimum frequency of >400MHz like in the old 3.4 kernel images would be even better.

MaxKellermann:
You don't need to recompile or change the kernel to change minimum/maximum frequencies; just set `/sys/devices/system/cpu/cpu*/cpufreq/scaling_min_freq`

tb59427:
@mihu - my idea (and I am currently working on it) is to add a separate "fix cpu freq" service to systemd (simple service - one time run) running a shell script with a set frequency (just like @MaxKellermann just commented). Currently trying to understand how I include that into the bitbake logic of OV....

Blaubart:
Can someone merge this and [this](https://github.com/Openvario/meta-openvario/pull/331) PR, please?

linuxianer99:
@tb59427 : There are conficts in the PR. Can you please fix this ?

tb59427:
Same situation as with the other commit: github says "Only those with write access to this repository can merge pull requests." And "resolve conflicts" is greyed out for me.

PR #331, opened 2022-09-20
Fix CPU speed (frequency) at 720MHz

tb59427:
During various tests with different settings for CPU voltage and CPU speed (see commit #0426cb7023a415d2c4645d19ec54f9462579bd5c
and commit #5cd841d1c151f8701aefecb393a72c0bd7ea82b5) it transpired that OV runs most stable with CPU speed fixed (thus stopping the ability to change processor speed). In my lab OV has run for more than two weeks in a row with this change.

I have implemented this as a small service run on (and on demand) by systemd. This
service invokes /opt/bin/fix_cpu_freq.sh - a small bash script fixing scaling_min and scaling_max frequencies for the CPU at the same value. Currently this value is 720MHz which has proven to be more than sufficient for my OV (5.7 no sensorboard).

If needed CPU speed can easily be changed in said script (see comments in the script).

One could argue that with this patch, my previous voltage patches (see commit numbers above) would be obsolete. True, if people stay with the 720MHz. If - for some reason - people change their min/max frequencies again, the voltage patch comes in handy as it ensures high enough voltage for the CPU for reasonable stability.


linuxianer99:
@tb59427 : There are conflicts ... Can you please fix ?

tb59427:
Dunno how, as github tells me "Only those with write access to this repository can merge files".

PR #334, opened 2022-10-30
OV stability fixes

tb59427:
This PR contains two commits:
**Further increase minimum CPU voltage to 1.2V** further increases CPU voltage to 1.2V. This has proven to increase 
stability greatly. However, it does not completely avoid OV freezes. It appears - at least on some OV systems - that changing CPU frequency occasionaly also creates freezes. 

**Fix CPU clock/frequency at 720MHz** fixes the issue of CPU clock changes occasionally creating freezes by fixing CPU 
clock at 720Mhz. See commit message, if you want to change the frequency to a different speed.

In combination these two fixes have created a rock solid OV (Stefly 5.7 LVDS) for me.


linuxianer99 (approved):

mihu-ov:
Finally :-)

Blaubart:
Thanks a lot!!

PR #338, opened 2023-03-11
Adding logbook to the new x-menu

Blaubart:
Logbook script to show last flights

This logbook script shows your last 20 flights in the X-menu at the OpenVario. It shows:

 -date
 -start time
 -landing time
 -flight duration


linuxianer99 (approved):

MaxKellermann:
This is useless code which will never be used by x-menu. Should be reverted.

linuxianer99:
How should we handle to ovmenu-ng vs. x-menu ?

I would suggest putting all effort to x-menu ...

PR #339, opened 2023-03-11
adding systeminfo script to the new x-menu

Blaubart:
This system info script shows your system configuration in the X-menu at the OpenVario. 
It provides information like:

- IP address of eth0
- MAC-address of eth0
- hostname
- IP address of waln0
- supplementary packages that are not included in every image: i2c-tools e2fsprogs usb-modeswitch
- Status of SSH, variod and sensord


linuxianer99 (commented):
How is this script called ??
Who calls it ??

Blaubart:
I changed title and description. It is for the new x-menu

MaxKellermann:
> It is for the new x-menu

This is useless for x-menu.

And this PR is formally wrong if it aims for the x-menu, because the script is in recipe `ovmenu-ng-skripts` which is for `ovmenu-ng` but not for `ovmenu-x`.

Everything about this PR is wrong.

Blaubart:
Why do think this is useless? It is also a shell script, but it only has to collect a couple of information and print them to the screen. So you think a shell script is not good enough to do this?

meta-ov/recipes-apps/xcsoar/files/system-info.sh
```diff
@@ -89,4 +89,4 @@ echo -e '\n'
 echo ' Status of SSH, variod and sensord:'
 echo ' SSH is '$SSH_STATUS
 echo ' variod is '$VARIOD_STATUS
-echo ' sensord is '$SENSORD_STATUS
+echo ' sensord is '$SENSORD_STATUS
```
lordfolken, line 92: missing newline at end of file. please fix your editor.

PR #340, opened 2023-03-14
moving logbook script to the right place

Blaubart:
The logbook script is part of the x-menu and not of ovmenu-ng. 

So it has to be placed at /meta-ov/recipes-apps/xcsoar/files instead of /meta-ov/recipes-apps/ovmenu-ng-skripts/files.


MaxKellermann:
This script is just as useless as this sysinfo script. It should never be used by x-menu.

Blaubart:
why do you think this script is useless?? The logbook is almost the same like the Kobo image uses it.

MaxKellermann:
Unlike the Kobo version, it outputs an ASCII text which works only with a fixed-width font, which obviously looks ugly. It is an obscure shell script that does very fragile things.
And it attempts to replicate code for which a better version already exists in the XCSoar repository.
The script is bad and redundant.

linuxianer99:
How should we handle to ovmenu-ng vs. x-menu ?

I would suggest putting all effort to x-menu ...

Blaubart:
I fully agree, too. I think we can close this PR?

PR #345, opened 2023-03-30
new daemon to store runtime values at config when shutdown the system

Blaubart:
This PR introduces a daemon that persists runtime variables during regular system shutdown. At the moment this daemon enables the rotation value to be saved in the /boot/config.uEnv in a reasonable manner, particularly with the X-menu. The daemon can easily be extended for more runtimer variables.

The daemon checks the current rotation setting in /sys/class/graphics/fbcon/rotate and store it in /boot/config.uEnv.

The PR also includes a change in ovmenu-ng.sh. Saving the display rotation in /boot/conf.uEnv has been removed here because this is now done by the daemon, primarily in preparation for the x-menu. Also the the display rotation is no longer saved in the /sys/class/graphics/fbcon/rotate_all, but in the /sys/class/graphics/fbcon/rotate. The reason for this is that the current rotation value can be read from the /sys/class/graphics/fbcon/rotate but not from /sys/class/graphics/fbcon/rotate_all.

So the daemon can be used with both the x-menu and the ng-menu. Its introduction later makes it easy to switch from the ng-menu to the x-menu.

### DIFF
--- a/meta-ov/recipes-apps/ovmenu-ng/files/ovmenu-ng.sh
+++ b/meta-ov/recipes-apps/ovmenu-ng/files/ovmenu-ng.sh
@@ -266,10 +266,8 @@ function submenu_rotation() {
 
 		 menuitem=$(<"${INPUT}")
 
-		# update config
 		# uboot rotation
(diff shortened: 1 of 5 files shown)


Blaubart:
no longer necessary!

PR #346, opened 2023-04-08
bug fix - sensord-testing was not build correctly

Blaubart:


linuxianer99 (approved):

MaxKellermann:
Bad commit message and very obscure diff. I have no idea what this does and what it fixes. I wouldn't have merged that.

PR #347, opened 2023-04-10
Update "update-maps.sh"

Blaubart:
If no ipk files and no xcm files are found, the script only returns two times the message to the user that no files for update were found. The message now states which file type was not found!


linuxianer99 (approved):

MaxKellermann:
Bad commit message

PR #350, opened 2023-04-25
Update Yocto, a syntax error fix and Kernel 6.2.12

MaxKellermann:


linuxianer99:
Sorry guys, but lots of work to do at the moment ... Will try to have a look on PR on the weekend ..

As requested by @MaxKellermann i should have a more closer look on the PRs, so this would take much longer ...

linuxianer99 (approved):

linuxianer99:
I spontaneously found time ;-)

MaxKellermann:
Lots of work, I know how that feels :-)

PR #320, opened 2022-07-21
improve backup and restore procedure

Blaubart:
This backup script stores all XCSoar settings and relevant OpenVario settings like:

- brightness of the display
- rotation
- touch screen calibration
- language settings
- dropbear settings
- SSH, variod and sensord status
 
backups are stored at USB stick at:
- openvario/backup/"MAC address of eth0"/
So you can store more than one backup on the same stick!

We add system info script, too

PR #352, opened 2023-04-29
Improve backup and restore procedure

7lima:
This is a revitalization of the pull requests #315, #317, and #320 which contains roughly the same outcome as those. Compared to them the separation into chunks/commits is different in a way that should make things clearer, especially with better consistency and coherence of these chunks.

The backup script stores all XCSoar settings and relevant OpenVario settings like:
    - brightness of the display
    - rotation
    - touch screen calibration
    - language settings
    - dropbear settings

Backups are stored on a USB stick at a place depending on the MAC address of eth0 which is unique for any device. So you can store more than one backup on the same stick.

We've added a reworked system info script, too. It gives an overview which packages and which versions are installed. It also shows IP addresses, kernel, host, system daemon status and other info. A bunch of new further system information was added also.

The backup and restore functions and menus have been made more user-friendly and should comprise a whole set of backup abilities now. Efficiency has been improved. Added a progress bar. Display error codes to the user transparently. Regarding the naming of functions and scripts consistency has been improved. This has been done by a rather complex renaming. Some old functionality has been dropped, new functionality has been added in the menus and backup scripts. See the individual commit comments for more details!

This is a joint work of @Blaubart and @7lima.


7lima:
This is more than ready for a merge. Any comments?

linuxianer99:
@Scumi Can you have a look please ? does it make sense, to include these improvements also in the new menu ?

Scumi:
Do you mean in the shell script menu? I suppose we can add it there until XCSoar is able to do this as well (OpenSoar can already as far as I see).

linuxianer99:
Yes, also fine for me. Just don't wanna loose some good ideas of the community.
Then i will close this PR and we will carry over the ideas to the XCSoar integrated menu, right ?

Scumi:
We could transfer it to a discussion to not lose it.

linuxianer99:
Also fine for me! Can you please do this ?

PR #354, opened 2023-05-15
Update XCSoar to 7.31

MaxKellermann:


lordfolken:
I think we should straight go to 7.33

PR #359, opened 2023-06-19
Restore sensord-testing service files installation.

### LINKED ISSUE #358: Sensord Testing not connecting to Variod
After updating all my repos to match master branch and performing a fresh full build, Sensord Testing stopped working.
It does not connect to Variod anymore. Variod is unable to find the Sensord socket. 
Sensord.socket file, Sensord@.service file and compdata file, necessary to compensate sensor drift, are missing.
I think the regression was introduced by #346, which removes the installation of these files.

iglesiasmarianod:
#346 removed the installation of sensord.socket and sensord@.service files needed by sensord testing to communicate with variod. As a result, communication between both daemons fails.
This PR restores the file to its previous state, prior to the regression.
Closes #358


MaxKellermann:
I don't get it.

"Delete unused sensord-testing service file and directory" - no, it's not unused. The commit message is misleading. The file is redundant and should never have been added (all of https://github.com/Openvario/meta-openvario/pull/346 was bad). But it's not unused!

"Clean and reorder sensord-testing recipe" - don't do two things in one commit! You reorder things, but what did you "clean"? It's very hard to see because it's obscured by the reorder. Why reorder, anyway?

" Add sensord-testing service files and remove obsolete sensord files from sensord-testing recipe" - this copies lots of code from `sensord.inc` into the -testing recipe. You pretend you authored this code, but you didn't. You created redundant copies of code that already exists, for no obvious reason and with no explanation.

Do not merge.

MaxKellermann:
btw. if #346 was a regression, then it should simply be reverted. I already commented over there that I didn't understand the PR, and never got a reply.

iglesiasmarianod:
Thanks @MaxKellermann for your time.
I never said I authored the code or pretend anything.
I only went back to where #284 left it, just before #346.
The redundant code was already there. 
I Didn't know that reverting #346 was possible either. 
If @linuxianer99 can revert the changes I'll close the PR.

linuxianer99:
Think reverting the #346 is possible. 

Let's revert #346 and adapt the sensord-testing recipe afterwards in regard to redundant code (sensord.inc).

Fine for all ??

iglesiasmarianod:
Thanks @linuxianer99 It's Ok for me.

iglesiasmarianod:
Hi @linuxianer99. I dropped my last three commits and reverted #346.
Please check if you think this is OK now.
Thanks.

linuxianer99 (approved):

Issue #375, opened 2024-05-22
OpenVario DS2.The reachable terrain line is not displayed.


arpass:
Hello friends,

I have encountered an unexpected problem with my OpenVario DS2. The reachable terrain line is not displayed. Additionally, the runways do not change color to green, although they do turn yellow. However, the required glide altitude is displayed correctly. The polar diagram for the Std.Libelle glider is activated. The flight altitude and speed display also work correctly. Could you please help me identify where to look for the issue?

Thank you in advance!

Scumi:
I can confirm those issues. It appears to be an issue with the UNIX target of XCSoar as the same issues appear on my linux laptop. We should open two bug reports on the XCSoar/XCSoar main repo.

Scumi:
So apparently an old compiler is responsible for the bug. Can we update the compiler used for the OpenVario build to its latest version (>= gcc-12/g++12)?

linuxianer99:
Should be fixed in upcoming releases ...

PR #376, opened 2024-06-09
replace defaulted equality operator to support pre g++-12 compiler

bomilkar:
This PR addresses https://github.com/XCSoar/XCSoar/issues/1253

g++ version 10.2 does not implement the default “operator==” construct in the same way as g++ version 12. This becomes clear if you replace the default “operator==” with the explicit “operator==” declaration: it solves Issue #375 and the issue 1253 reported against XCSoar.

Defaulted equality comparison, a C++-20 feature, is used in XCSoar:
src/Geo/Flat/FlatGeoPoint.hpp
This is addressed in the first commit.

There are other files using defaulted “operator==” constructs:
src/Math/Point2D.hpp
src/Geo/GeoPoint.hpp
src/PageSettings.hpp
Instances in these 3 files have not been identified as root causes to other Issues. I still recommend to apply the patches just to be safe. These changes are in a separate commit.

This PR will be obsolete as soon at the OV build deploys compiler version g++-12 (or later).

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,17 @@ This will fetch the manifest file for the warrior branch.
 
 ### Starting the containerd build environment
 
+The "old" Ubuntu 18.04 build container (for warrior branch):
+
 ```
 sudo docker run -it --rm -v $(pwd):/workdir linuxianer99/ovbuild --workdir=/workdir
+cd poky
+```
+
+The "new" Debian 11 build container (for hardknott branch):
+
+```
+sudo docker run -it --rm -v $(pwd):/workdir ghcr.io/openvario/ovbuild-container:main --workdir=/workdir
 cd poky
 ```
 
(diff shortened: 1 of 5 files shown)


MaxKellermann:
OpenVario is based on Yocto 4.0 Kirkstone which comes with GCC 11, not 10.
But a sane way of dealing with compiler bugs is to apply compiler bug fixes instead of making the XCSoar code worse. Even if GCC 11 were still affected from this compiler bug (which you did not claim), it would be easy to update the compiler, e.g. by updating to a later Yocto version. Yocto 4.0 is two years old already; in October 2022, Yocto 4.1 was released with GCC 12, and two months, Yocto 5.0 was released with GCC 13.

Scumi:
We were not successful yet updating to a later Yocto version (the image does not boot). Might be a winter project as I don't have my OpenVario at home. We haven't thought of fixing the compiler directly, but I agree this would be the better way (if possible with my programming skills).
I can also confirm that the bug is present with GCC 11.

bomilkar:
The symptoms have been described in several Issues here and in XCSoar. One of them is mentioned in this PR.
We narrowed the cause down to the implementation of the defaulted equality operator (be it GCC10 or 11).
BTW: declaring the inequality operator along side with the equality operator doesn't seem to be covered by the LRM: https://en.cppreference.com/w/cpp/language/default_comparisons
It's unclear what happens if both are declared.

PR #383, opened 2024-11-19
Add SD Card Image as build artifact

linuxianer99:


mihu-ov:
Great to see progress on the OV image, thank you!

>    The disk layout used is:
>    0                      -> IMAGE_ROOTFS_ALIGNMENT         - reserved for other data
>    IMAGE_ROOTFS_ALIGNMENT -> BOOT_SPACE                     - bootloader and kernel
>    BOOT_SPACE             -> SDIMG_SIZE                     - rootfs__

Would this be the right time to also implement an extra data partition?

linuxianer99:
Yes this would be needed ...
Currently we are thinking about using RAUC as a update tool (https://rauc.io/).
This supports also data partitions which will survive a complete flash process ...

Maybe you can also have a look on RAUC ??

Issue #396, opened 2026-01-05
RPi CM4 integration


Maxel01:
@markus2marks has been working on integrating the Raspberry Pi CM4 for the OpenVario. The current configuration does not provide a valid integration of the new platform. After speaking with Markus, I would like to integrate his changes step by step into the newest branch.

My plan is to take a look at every commit he has pushed in his [fork](https://github.com/markus2marks/meta-openvario/tree/hardware_bringup) and create PRs for them.

The PRs will have commits beginning with `Needs Investigation:`. These commits should be checked before merging as they contain code changes that I cannot check if everything is still working. I'll change the commit messages after the commits have been approved. Is that a way to go?

---

This is an overview of the upcoming tasks:

#### Needs decision:
- [ ] [add dts files for 7 inch displays](https://github.com/markus2marks/meta-openvario/commit/c36e21a4841f55a4a3f449744a60de0f4d2accec) (implemented in [eadf5c0](https://github.com/Openvario/meta-openvario/commit/eadf5c0c1dcb23a02f53ae393a17c4e1bae58a2a) as `pq070` instead of `lvds` and `lvds-ampire`)
- [ ] [remove meta-yocto](https://github.com/markus2marks/meta-openvario/commit/f82d4dd476eefbb7456c3c792fa5b600b60c690f): already removed?
- [ ] [remove unsed submodules](https://github.com/markus2marks/meta-openvario/commit/df60f3f8abac20d670e392ad720174f6299ed19e): already removed?
- [ ] [update submodules](https://github.com/markus2marks/meta-openvario/commit/faf646a8abaa398ada04af1614d02468cad37ddc) what needs to be updated?

#### To-Do:
- [ ] [add rauc + ~~wlan~~](https://github.com/markus2marks/meta-openvario/commit/9d44a90d4cf53275fc5acccfdffb63845b7ce023)
- [ ] [update dts and conf files](https://github.com/markus2marks/meta-openvario/commit/04c591a8ec9453e693cabde1b3184c19fb33eb70)
- [ ] [add submodule rauc, add script for img build, add wpa_supplicant-wlan0.conf file](https://github.com/markus2marks/meta-openvario/commit/b905291d39f8d2362e66923d69f9146353e77705)
- [ ] [data partition is generated correctly](https://github.com/markus2marks/meta-openvario/commit/d496c8cf0a361ff17c2d6e5b6ce0b8a955de0df0)
[...]

PR #397, opened 2026-01-05
CM4: change screen rotation, activate i2c

Maxel01:
Adding I2C to enable communication with the sensorboard

The XCSoar version needs to be updated to integrate the fix for the CM4 platform for issue Openvario/meta-openvario#387.

The rotation part might be a change for the CM4 platform but could affect the cubieboard build: any comments?

Move otg_mode from bootfiles to machine configs

Part of #396

### DIFF
--- a/conf/machine/ov-cm4-57-lvds.conf
+++ b/conf/machine/ov-cm4-57-lvds.conf
@@ -36,6 +36,8 @@ UBOOT_MACHINE = "rpi_arm64_config"
 #Enable u-boot
 RPI_USE_U_BOOT = "1"
 ENABLE_UART = "1"
+#Enable otg mode
+ENABLE_DWC2_HOST = "1"
 
 VC4DTBO ?= "vc4-kms-v3d"
 
--- a/conf/machine/ov-cm4-7-pq070.conf
+++ b/conf/machine/ov-cm4-7-pq070.conf
@@ -36,6 +36,8 @@ UBOOT_MACHINE = "rpi_arm64_config"
 #Enable u-boot
 RPI_USE_U_BOOT = "1"
 ENABLE_UART = "1"
+#Enable otg mode
+ENABLE_DWC2_HOST = "1"
 
 VC4DTBO ?= "vc4-kms-v3d"
 
--- a/conf/machine/ov-rpi4-64.conf
+++ b/conf/machine/ov-rpi4-64.conf
@@ -36,6 +36,8 @@ UBOOT_MACHINE = "rpi_arm64_config"
 #Enable u-boot
 RPI_USE_U_BOOT = "1"
 ENABLE_UART = "1"
+#Enable otg mode
+ENABLE_DWC2_HOST = "1"
 
 VC4DTBO ?= "vc4-kms-v3d"
 
--- a/conf/machine/ov-rpi4.conf
+++ b/conf/machine/ov-rpi4.conf
@@ -24,6 +24,8 @@ RPI_KERNEL_DEVICETREE ?= " \
 # Enable U-boot
 RPI_USE_U_BOOT = "1"
 ENABLE_UART = "1"
+#Enable otg mode
+ENABLE_DWC2_HOST = "1"
 
 # 'l' stands for LPAE
 SDIMG_KERNELIMAGE ?= "kernel7l.img"
--- a/recipes-apps/ovmenu-ng/files/ovmenu-ng-rpi.sh
+++ b/recipes-apps/ovmenu-ng/files/ovmenu-ng-rpi.sh
@@ -1,0 +1,434 @@
+#!/bin/bash
+
+#Config
+TIMEOUT=3
+INPUT=/tmp/menu.sh.$$
+
+# trap and delete temp files
+trap "rm $INPUT;rm /tmp/tail.$$; exit" SIGHUP SIGINT SIGTERM
+
+main_menu () {
+while true
+do
+	### display main menu ###
+	dialog --clear --nocancel --backtitle "OpenVario" \
+	--title "[ M A I N - M E N U ]" \
+	--begin 3 4 \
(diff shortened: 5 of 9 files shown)


Scumi:
XCSoar 7.44 is not officially out yet, I'd wait with that. However, it seems not far away, looking at the current discussions.

Rotation should work for cubieboard, there are still a few around. Can we build ovmenu-ng.sh based on machine? 
My long term goal would anyway be to have an XCSoar based main menu and not rely on this script anymore.

Maxel01:
Another option could be to define the XCSoar version depending on the machine. It doesn't make sense to use the old commit when the version is not even running on CM4.

I am currently working on creating a separate menu (so building different ones for cubieboard and for CM4). It would be good to implement the menu in XCSoar. I'd like to prepare the CM4 image for the season, so going with the current main menu.

I'll change these two things.

Scumi:
I agree in general, but updating to 7.44 might be misleading because its not officially out yet. Could using the -testing build for building from master branch work? In any case, I do not expect 7.44 NOT to be out by start of the season (@lordfolken ?).

Maxel01:
The idea behind the XCSoar upgrade (commit from January 2025) was to avoid bad builds for the raspberry pi platform. It is annoying starting with a project and getting problems that you need to fix yourself. That's what I had to do yesterday. That's the reason why I proposed to merge Markus' changes. Maybe we can agree on my current solution? If not I'll revert that until 7.44 is ready.

I managed to add a custom main menu for the raspberry pi which will be extended in the next PRs.

linuxianer99:
Hey Maxel,
thanks for the commits. I guess all these things are already in the code but in a different branch.
I did a prerelease for the RPI with Markus.

Maybe you can have a look here:

https://github.com/linuxianer99/meta-openvario/tree/hardware_bringup

Would also be great if we can have just a short email or phone conversation to align on the strategy. So we can avoid double work. What do you thing ?
You can just give me a short notice on openvario@timoshome.org

Thanks :-)

Scumi:
> The idea behind the XCSoar upgrade (commit from January 2025) was to avoid bad builds for the raspberry pi platform. It is annoying starting with a project and getting problems that you need to fix yourself. That's what I had to do yesterday. That's the reason why I proposed to merge Markus' changes. Maybe we can agree on my current solution? If not I'll revert that until 7.44 is ready.
> 
> I managed to add a custom main menu for the raspberry pi which will be extended in the next PRs.

I still don't understand it fully. Does the testing branch not build for you? It should pull bleeding edge 7.44 from XCSoar so it should run on RPi4? I try to find an RPi4 for myself to check and help. My OpenVario happily still runs on the Cubieboard :-)

linuxianer99 (approved):
See my comments below ... sorry ... wrong button click ;-)

linuxianer99 (commented):
I would prefer not to set the XCSoar version based on machine ...
Why it is so important to have this prerelease 7.44 in the CM4 Build ?

lordfolken:
Can i make an ipkg feed for xcsoar? Baking it into an image isnt that practical as you need to reflash and restore each time.

linuxianer99:
> Can i make an ipkg feed for xcsoar? Baking it into an image isnt that practical as you need to reflash and restore each time.

For sure ! Question is where should we store the feed .... let me have a look ...

Maxel01:
> Would also be great if we can have just a short email or phone conversation to align on the strategy. So we can avoid double work. What do you thing ?

@linuxianer99 I'll write you a message.

> thanks for the commits. I guess all these things are already in the code but in a different branch. I did a prerelease for the RPI with Markus.
> 
> Maybe you can have a look here:
> 
> https://github.com/linuxianer99/meta-openvario/tree/hardware_bringup

I think it does work with the changes in the branch but it is not complete. The I2C has been activated in [Enable I2C and LVDS Bridge](https://github.com/linuxianer99/meta-openvario/commit/e6e5767bc583b4d3282f31c650c0fea2c20404ef) but has been removed again in [Remove XCSoar for testing and add development image](https://github.com/linuxianer99/meta-openvario/commit/6af605e571042d91e19129b510a5e0bad5214f3d). The rotation setting hasn't been updated, wifi and bt is not yet implemented. Markus did some work on that. Additionally with the new structure, these branches cannot be merged directly.

> I still don't understand it fully. Does the testing branch not build for you? It should pull bleeding edge 7.44 from XCSoar so it should run on RPi4? I try to find an RPi4 for myself to check and help. My OpenVario happily still runs on the Cubieboard :-)

The build is not the problem. XCSoar does not start. It has something to do with basic_string_view. This has been discussed in #387 and updated in XCSoar in XCSoar/XCSoar#1637 which is newer than the 7.43 release. I did not try the testing image yet because the llvm build does not succeed on my laptop (even docker crashes, not enough RAM) and I currently do not have access to my other computer where it works.

linuxianer99:
> > Can i make an ipkg feed for xcsoar? Baking it into an image isnt that practical as you need to reflash and restore each time.
> 
> For sure ! Question is where should we store the feed .... let me have a look ...

@lordfolken I will contact Andreas (owner of www.openvario.org) if we can host the feed there. On the other side i have claimed openvario.de a well as openvario-flightcomputer.de in the afternoon. So we should have domains for hosting the feed ;-)

linuxianer99 (changes requested):
Can you please use the "xcsoar-testing.bb" recipe for the rpi build ?
Think the recipe "xcsoar_7.43.bb" should compile the release tags 7.43 of XCSoar

Maxel01:
> Can you please use the "xcsoar-testing.bb" recipe for the rpi build ?
> Think the recipe "xcsoar_7.43.bb" should compile the release tags 7.43 of XCSoar

@linuxianer99 I removed the commit. I'll use the testing image until 7.44 is released.

linuxianer99 (approved):

PR #400, opened 2026-01-10
RPi4: move common configuration to top-level machine

Maxel01:
With the new RPi platform 4 new machines have been created with no top-level machine. The configurations are almost the same. That's why it would make sense to create a top-level setup with all common configurations.

The `ov-rpi4.conf` has been adjusted to match the current `ov-rpi4-64.conf`. The differences were introduced in [Enable u-boot for raspberry pi](https://github.com/Openvario/meta-openvario/commit/8584038dfd30895cb798249fab9645d0a5c39f4f) and [Corrected machine settings](https://github.com/Openvario/meta-openvario/commit/d746ba6eb3fb690f5e0eeb5620d2c5e28fc9dd34).

I don't have a RaspberryPi 4 (32bit), so maybe someone could verify that it is still working on the Raspberry Pi.

### DIFF
--- a/conf/machine/ov-cm4-57-lvds.conf
+++ b/conf/machine/ov-cm4-57-lvds.conf
@@ -2,54 +2,17 @@
 #@NAME: RaspberryPi 4 Development Board (64bit)
 #@DESCRIPTION: Machine configuration for the RaspberryPi 4 in 64 bits mode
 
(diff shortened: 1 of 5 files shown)


linuxianer99 (approved):
I am currently in the phase of changing machine names anyway .. so we have to adapt again .. But lets do this step by step ...

PR #415, opened 2026-03-18
Fix/misc fixes

linuxianer99:
Misc fixes/improvements:
- Add missing files (data partition)
- Introduce  DISTRO_FEATURE for selecting Main Application (Opensoar vs. XCSoar)
- Adapt package feed
- Adapt ovmenu_ng to new version of busybox

### DIFF
--- a/conf/distro/ovlinux.conf
+++ b/conf/distro/ovlinux.conf
@@ -12,7 +12,7 @@ SDK_VERSION := "${@'${DISTRO_VERSION}'.replace('snapshot-${DATE}','snapshot')}"
 MAINTAINER = "Timo Bruderek <openvario@timoshome.org>"
 
 # Image Version suffix constructed from git tags and commit info
-IMAGE_NAME = "OV-${IMAGE_VERSION_SUFFIX}-${SHORT_OV_MACHINE}"
+IMAGE_NAME = "OV-${IMAGE_VERSION_SUFFIX}-${SHORT_OV_MACHINE}-${IMAGE_TYPE}"
 #IMAGE_NAME_LINK = "OV-${IMAGE_BASENAME}-glibc-ipk-current-${MACHINE}"
 
 TARGET_VENDOR = "-ovlinux"
@@ -49,7 +49,7 @@ DEPENDS:remove:pn-libsndfile1 = "flac libogg libvorbis"
 EXTRA_OECONF:append:pn-libsndfile1 = " --disable-external-libs"
 
 PACKAGECONFIG:remove:pn-libxml2:cubieboard2 = "python"
-PACKAGECONFIG:remove:pn-libxml2-native:cubieboard2 = "python"
+PACKAGECONFIG:remove:pn-libxml2-native = "python"
 
 PACKAGECONFIG:remove:pn-alsa-plugins = "samplerate speexdsp"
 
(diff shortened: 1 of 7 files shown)


Maxel01:
I will take a look at it in the next few days.

conf/distro/ovlinux.conf
```diff
@@ -94,6 +94,7 @@ DISTRO_FEATURES:append = " systemd usrmerge"
 DISTRO_FEATURES:remove = "acl argp pcmcia usbgadget xattr nfs zeroconf pci 3g nfc x11 sysvinit"
 DISTRO_FEATURES:remove:cubieboard2 = "gobject-introspection-data"
 DISTRO_FEATURES:append:raspberrypi4 = " bluez5 bluetooth wifi glib systemd-resolved"
+DISTRO_FEATURES:append = " opensoar"
```
Maxel01, line 97: I assume that the decision to use Opensoar as default has been made, right?

recipes-core/images/openvario-image-testing.bb
```diff
@@ -15,7 +14,10 @@ IMAGE_INSTALL += "\
     ovmenu-ng \
 "
 
+IMAGE_INSTALL:append = "${@bb.utils.contains('DISTRO_FEATURES', 'opensoar', ' opensoar-testing', '', d)}"
```
Maxel01: I already had the issue before.
Is there a opensoar-testing recipe? I am getting "no buildable providers"
I also get a new error:
```
bb.data_smart.ExpansionError: Failure expanding variable IMAGE_INSTALL, expression was ... ${@bb.utils.contains('DISTRO_FEATURES', 'opensoar', '', d)} which triggered exception TypeError: contains() missing 1 required positional argument: 'd'
```
It is also inconsistent and includes both xcsoar and opensoar when selecting opensoar. What is the reason, that opensoar and xcsoar shouldn't be enabled at the same time (in the release image)?
Also consistency across all available images

recipes-core/images/openvario-image.bb
```diff
@@ -19,5 +28,7 @@ IMAGE_INSTALL += "\
 "
 
 IMAGE_INSTALL:append = "${@bb.utils.contains('DISTRO_FEATURES', 'opensoar', ' opensoar', '', d)}"
+IMAGE_INSTALL:append = "${@bb.utils.contains('DISTRO_FEATURES', 'xcsoar', ' xcsoar', '', d)}"
 
 export IMAGE_BASENAME = "openvario-image"
+export IMAGE_TYPE = ""
```
Maxel01: I am not familiar with the other build options (openvario-image-develop, ...).
These will fail without setting the `IMAGE_TYPE`

recipes-core/images/openvario-image-testing.bb
```diff
@@ -15,7 +14,10 @@ IMAGE_INSTALL += "\
...
+
 export IMAGE_BASENAME = "openvario-testing"
+export IMAGE_TYPE = "testing"
```
Maxel01, line 19: I am not familiar with the other build options (openvario-image-develop, ...).
These will fail without setting the `IMAGE_TYPE`

conf/distro/ovlinux.conf
linuxianer99: Right! But it can also be changed to XCSoar ...

recipes-core/images/openvario-image-testing.bb
linuxianer99: The reason is the ovmenu selector. There a variable main_app set in uboot to select which App will be started be ovmenu-ng. So the both binary can exists, but one has to be selected.

I removed the IMAGE_INSTALL line in testing image and add opensoar always to the image. There is also no testing version.

recipes-core/images/openvario-image-testing.bb
linuxianer99: The develop is a real real development option ... Never used this is up to the developer to use this ... but never for any release ..

recipes-core/images/openvario-image-testing.bb
Maxel01: I can't remember from yesterday but did it fully build for you with the IMAGE_INSTALL in the release openvario-image? Maybe there is something wrong with my build environment.

Maxel01: I'll approve it, but check it this evening or tomorrow.

Maxel01 (approved):

PR #413, opened 2026-03-11
Add xcsoar-locale to cm4 build

Maxel01:
In the CM4 build, the selected xcsoar language is not available in the file system. I assume that everything works just fine with the cubieboard build.
I added all selected language defined in openvario-base-image.
I wanted to add it to the raspberrypi4 override only but that did not work:
```
IMAGE_INSTALL:raspberrypi4 += " \
    ${XCSOAR_LOCALE_PACKAGES} \
"
```
Everything before is removed. That's the result with the override.
```
IMAGE_INSTALL="           xcsoar-locale-de     xcsoar-locale-es     xcsoar-locale-fr     xcsoar-locale-it     xcsoar-locale-hu     xcsoar-locale-pl     xcsoar-locale-cs     xcsoar-locale-sk     xcsoar-locale-ru     xcsoar-locale-lt      xcsoar-testing     xcsoar-menu     xcsoar-profiles     xcsoar-maps-default     sensord-testing    variod-testing     ovmenu-ng  connman connman-client     udev-rules-rpi "
```
That's the expected and current output:
```
IMAGE_INSTALL="     packagegroup-base     distro-feed-configs     nano     openssh-sftp-server     tslib     tslib-tests     tslib-uinput     tslib-conf     tslib-calibrate     touch-udev-rules     ts-uinput-service          linux-firmware-ralink     linux-firmware-rtl8168     linux-firmware-rtl8188     linux-firmware-rtl8192     linux-firmware-rtl8712     linux-firmware-rtl8723     linux-firmware-rtl8821           locale-base-en-us     locale-base-de-de     locale-base-es-es     locale-base-fr-fr     locale-base-it-it     locale-base-hu-hu     locale-base-pl-pl     locale-base-cs-cz     locale-base-sk-sk     locale-base-ru-ru     locale-base-lt-lt      fix-cpu-freq     xcsoar-testing     xcsoar-menu     xcsoar-profiles     xcsoar-maps-default     sensord-testing    variod-testing     ovmenu-ng          xcsoar-locale-de     xcsoar-locale-es     xcsoar-locale-fr     xcsoar-locale-it     xcsoar-locale-hu     xcsoar-locale-pl     xcsoar-locale-cs     xcsoar-locale-sk     xcsoar-locale-ru     xcsoar-locale-lt   connman connman-client     udev-rules-rpi "
```
Any Ideas?
This also needs to be added to the standard image.
Any Ideas to avoid duplication?

PR #416, opened 2026-03-19
xcsoar: include locale files in main package

Scumi:
A proposal that should fix the language-setting issue on the xcsoar recipe level. 

Other option: The fix of #413 would need to be considered on every image recipe and kept in sync with potentially added/removed locales in XCSoar itself.

### DIFF
--- a/meta-ov/recipes-apps/xcsoar/xcsoar.inc
+++ b/meta-ov/recipes-apps/xcsoar/xcsoar.inc
@@ -36,6 +36,10 @@ S = "${WORKDIR}/git"
 
 LC_LOCALE_PATH = "${datadir}/locale"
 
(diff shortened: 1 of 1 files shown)


Maxel01:
Looks good to me. It is definitely a better solution than #413.
It needs to be included in the scartgap branch too.

linuxianer99:
So we need both ?? #413 and #416 ?

Maxel01:
I don't think so. This is the better solution. I will later check the issue with the xcsoar translation change in the infoboxes only after restart.

Scumi:
I only compiled with this, as alternative proposal to #413.

linuxianer99:
ok. Then i will merge #416 and we close #413 .

Maxel01:
@linuxianer99 you can merge this PR.

PR #418, opened 2026-03-23
xcsoar: include locale files in main package

Maxel01:
Same as #416 but for the current scarthgap branch.


linuxianer99 (approved):

linuxianer99:
@Maxel01 Thanks for porting it!

PR #419, opened 2026-03-24
xcsoar: update to 7.44

Scumi:
I'd also update the scarthgap branch, but I'm not entirely sure yet how the new build process works.

### DIFF
--- a/meta-ov/recipes-apps/xcsoar/xcsoar.inc
+++ b/meta-ov/recipes-apps/xcsoar/xcsoar.inc
@@ -42,8 +42,7 @@ PACKAGES:remove = "${PN}-locale"
 
 SRC_URI += " \
 	file://0007-Disable-touch-screen-auto-detection.patch \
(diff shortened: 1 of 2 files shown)


Maxel01:
I'm not quite sure as I've never used the previous build process but I assume it is pretty much the same. The build structure has been moved to [Openvario/OpenVario](https://github.com/Openvario/OpenVario) which means that this repo only contains the meta-openvario (meta-ov) directory. 
That should be it (I hope, at least the instructions are the same).
I already prepared something for the scarthgap branch. I'll create the PR now.

Scumi:
I see, must have missed that one. Thanks.

linuxianer99:
@Scumi Sorry for that. I tried to change a lot of things in the last months, but it is not fully documented yet.

I updated the Readme in the Openvario Repository. Should the instructions are fine now. (Only change was container image version)

Scumi:
All good, I figured it out. We can only work with the time we have :-)

PR #409, opened 2026-03-01
CM4: Add Wifi Config and Extend Main Menu

Maxel01:
# Add license flags to successfully build

This PR adjusts recipe/packaging settings by introducing the necessary license flags. This change was required to make the project build cleanly without license‑related errors. (see config change in [2892df0](https://github.com/Openvario/meta-openvario/commit/2892df05a9eaff75097f968784c226e1a4dbd2a5))

# Add wifi config and extend main menu

It introduces a new Wi‑Fi configuration component and expands the main menu interface to expose this functionality.

Together these updates prepare the build environment for success and add basic wireless configuration support via an enhanced menu.

PR #421, opened 2026-03-25
CM4: remove low lcd brightness levels

Maxel01:
Removes very dark brightness levels. They were not even usable in dark environments.
However I encountered the following **issue**:
High pitch tone increasing for brightness levels from level 1 to 4 (already starting with a quiet tone at level 1) and decreasing to level 10. At level 9 it is still hearable, at level 10 the high pitch tone is gone.

Previous behavior (high pitch tone):
increasing: level 6 to 9
decreasing: level 9 to 10
levels 11 to 15 not selectable.

In addition the brightness level is not stored correctly (after reboot the default value is loaded and present in the config file). Does anyone know how it is handled on the cubieboard or what the problem could be?

### DIFF
--- a/recipes-kernel/linux/files/rpi/ov-cm4-ch57-overlay.dts
+++ b/recipes-kernel/linux/files/rpi/ov-cm4-ch57-overlay.dts
@@ -39,8 +39,8 @@
         
             lcd:lcd {
                 compatible = "pwm-backlight";
-                brightness-levels = <0 6 8 12 16 24 32 40 48 64 96 128 160 192 224 255>;
-                default-brightness-level = <15>;
+                brightness-levels = <24 32 40 48 64 96 128 160 192 224 255>;
+                default-brightness-level = <10>;
                 pwms = <&pwm 0 200000 0>;
             };
 
--- a/recipes-kernel/linux/files/rpi/ov-cm4-pq70-overlay.dts
+++ b/recipes-kernel/linux/files/rpi/ov-cm4-pq70-overlay.dts
@@ -39,8 +39,8 @@
         
             lcd: lcd {
                 compatible = "pwm-backlight";
-                brightness-levels = <0 6 8 12 16 24 32 40 48 64 96 128 160 192 224 255>;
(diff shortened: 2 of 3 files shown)


Maxel01:
I tested against this commit [85957bd](https://github.com/Openvario/meta-openvario/commit/8b957bd9480a3b8cfaa4ec4056342a0e1c25b6da) and [62927c](https://github.com/Openvario/meta-openvario/commit/62927cabc95e3cfb6ca06835f9d5672c4933164a), #404 (both with 57-lvds). The display brightness is restored correctly.
Apparently it changed with #409.
I assume it is this line in `rpi-cmdline.bbappend`:
```
CMDLINE:append = " systemd.restore_state=0 rfkill.default_state=1"
```

linuxianer99 (approved):

Maxel01:
@markus2marks I checked with the newest scarthgap commit und removed `systemd.restore_state=0`. Now the brightness level will be restored. Do you know how to solve that problem?

linuxianer99:
No idea. Think we have to dig deeper into systemd ...

Maxel01:
I don't quite know why Markus added it, so maybe he can say something to that.

markus2marks:
I have no idea either. Also, we should reset the PWM frequency to the default setting, just like on the Cubieboard. That should make the noise go away. I used an online template for that and didn't adjust the frequency any further.

markus2marks:
I checked my file, but I can't find that line: “systemd.restore_state=0 rfkill.default_state=1” in my cmdline.txt. That's why it works for me.

Maxel01:
Mhh, in recipes-bsp/bootfiles/rpi-cmdline.bbappend is the line in the hardware_bringup branch. Could you check that?

If it is not necessary I will remove that.

Maxel01:
> I have no idea either. Also, we should reset the PWM frequency to the default setting, just like on the Cubieboard. That should make the noise go away. I used an online template for that and didn't adjust the frequency any further.

Could you maybe do that?

markus2marks:
yes, i will check that

PR #422, opened 2026-03-28
Fix/split menus

Scumi:
As far as I see, the menu updates of 51728fadfc7ca75d69b981fb63eadf8088bf29bf only work for OpenSoar right now. 
`ovmenug-ng-rpi.sh` however appears to be the old menu, plus WiFi via NetworkManager. 

This PR basically splits the menus and installs them based on the `DISTRO_FEATURES` selection for OpenSoar or XCSoar. The networking submenu is disabled on Cubieboard for now by probing the availability of NetworkManager.


Maxel01:
First of all, I created the separate menu for the rpi because I didn't know how to solve the problem properly and it would be easier to add functionality specific for the rpi (such as FTP server, bluetooth connections etc). So it's perfect if there is a better way for that.

With the opensoar merge the menu changed and I didn't have time to check if the rpi menu still works as expected. I assume that some functionality has been broken by the removal of some shell script files. You can check them in the same commit (e.g. Update_Maps.sh has been removed and is still used in the previous / rpi menu).

There is an ongoing discussion about the integration of the the ov-menu inside xcsoar/opensoar.
Should we maybe create a separate issue for that? At the moment the discussion is spread across several issues and (at least for me) there is no clear line for the future.

Scumi:
> First of all, I created the separate menu for the rpi because I didn't know how to solve the problem properly and it would be easier to add functionality specific for the rpi (such as FTP server, bluetooth connections etc). So it's perfect if there is a better way for that.

What I don’t fully understand is why additional software is being installed on the RPi in the first place. Wouldn’t something like connman already cover the requirements? I’d also like to switch to NetworkManager on the Cubieboard, but that risks breaking setups for users with custom menus (something we’d need to address separately). In general, diverging configurations will make this harder to maintain for everyone.

Edit: I see that display rotation might be different as well, not only software.

> With the opensoar merge the menu changed and I didn't have time to check if the rpi menu still works as expected. I assume that some functionality has been broken by the removal of some shell script files. You can check them in the same commit (e.g. Update_Maps.sh has been removed and is still used in the previous / rpi menu).

Thanks for pointing this out. I adopt the menu accordingly.

> There is an ongoing discussion about the integration of the the ov-menu inside xcsoar/opensoar.
> Should we maybe create a separate issue for that? At the moment the discussion is spread across several issues and (at least for me) there is no clear line for the future.

I know, I created that [discussion](https://github.com/Openvario/meta-openvario/discussions/401) (though its been ongoing for a long while) :-) I am on it, but it takes time. Until then, there needs to be a working menu.

Funnily enough, the menu items that do not work right now (due to renamed or missing scripts) are the ones that soon might not be needed anymore because extended file management is almost ready from inside of XCSoar (https://github.com/XCSoar/XCSoar/pull/2158).

Maxel01:
> What I don’t fully understand is why additional software is being installed on the RPi in the first place. Wouldn’t something like connman already cover the requirements?

Well, Markus added that in his hardware_bringup branch. I ported these changes to the scarthgap branch. I think, we could also move back to connman.

> I’d also like to switch to NetworkManager on the Cubieboard, but that risks breaking setups for users with custom menus (something we’d need to address separately). In general, diverging configurations will make this harder to maintain for everyone.

I agree. When you say you want to move to NetworkManager why? Is there a reason (besides copy&paste on my side)?

> Funnily enough, the menu items that do not work right now (due to renamed or missing scripts) are the ones that soon might not be needed anymore because extended file management is almost ready from inside of XCSoar (XCSoar/XCSoar#2158).

Nice, but these files need to be readded until it's done, right?

> Edit: I see that display rotation might be different as well, not only software.

Yes... maybe that was the reason, why created the new file. As you said, it might be beneficial to have only one menu (maybe splitting for opensoar and xcsoar). Is there a way to integrate the rotation using an external shell script and include the correct one during the build?

Scumi:
> Well, Markus added that in his hardware_bringup branch. I ported these changes to the scarthgap branch. I think, we could also move back to connman.

Got it. I’m currently struggling to see a clear direction for some of the changes (though I might be missing something). The reasoning isn’t always obvious from the commit messages or pull requests.

> I agree. When you say you want to move to NetworkManager why? Is there a reason (besides copy&paste on my side)?

Using NetworkManager could help integrate network management directly into XCSoar.
Alternatively, we could stick with a simpler setup like wpa_supplicant + DHCP, similar to KOBO/OpenSoar.

NetworkManager has the advantage that it’s widely available and testable on most desktop Linux systems. However, even in a stripped-down configuration, it introduces more overhead compared to a minimal wpa_supplicant-based solution.

> Nice, but these files need to be readded until it's done, right?

Maybe, looking into that right now. We might be able to rely on `transfer-xcsoar.sh` instead.

> Yes... maybe that was the reason, why created the new file. As you said, it might be beneficial to have only one menu (maybe splitting for opensoar and xcsoar). Is there a way to integrate the rotation using an external shell script and include the correct one during the build?

I’d probably detect the available backend at runtime and reuse a single menu, rather than splitting things further.

linuxianer99 (approved):

Scumi:
I should have marked it WIP, it needs some more fixes. I will create a follow-up PR the next days that should make it work for all current configurations.

PR #425, opened 2026-04-01
Fix/menu and xcsoar recipes

Scumi:
follow-up to #422.

Should get us back to the old menu for XCSoar (also working on RPi) until the integrated menu similar to OpenSoar is ready.

@Maxel01: Please check rotation on your end please.


recipes-apps/xcsoar/xcsoar_7.43.bb
Maxel01, line 1: Could you also update or remove  the preferred version I added in #420 
It has been added because I didn't know if the XCSoar release is official now as the first release has been withdrawn.

Scumi: Good catch, will do!

Maxel01:
Yes, the rotation is applied and restored on boot up.

linuxianer99:
@Scumi Does this also work in combination with Opensoar ?? 
Or should we also set the "main_app" to XCSoar as default ?

Scumi:
There should be no regressions on OpenSoar. `main_app` is set on `config.uEnv` during compilation depending on which app is set in `DISTRO_FEATURES`.

linuxianer99:
ok. So i will merge ?

Scumi:
Yes, I would think so.

Maxel01:
Opensoar does automatically start on the RPi but that has nothing to do with these changes.
With a manual start via the terminal with `/usr/bin/OpenSoar` opensoar opens as intended.
The first message I get (and that I can read before disappearing):
```
Call another ovmenu-ng.sh to change it 'on the fly'
```

Before #422 the opensoar menu opens but opensoar is only startable from the terminal. At least the ovmenu opens there. That's not happening with this PR and the changes made in #422 
@Scumi did you try it with the Opensoar build? Does it work or are there any issues?

Scumi:
Interesting, I didn't observe this. Will check again, thanks.

Scumi:
So I tried a build with and without changes of this PR and OpenSoar works exactly the same - no regression observed. Can be merged from my point of view.

Issue #391, opened 2025-03-27
New Openvario Image / Yocto Scarthgap?


mihu-ov:
Are there any plans to release current Openvario images for the start of gliding season in the northern hemisphere?
I was able to compile current master but failed with master-next.

Scumi:
My master-next compiles just fine. I think you have to pull the latest docker image.

I fully support a new release and would create a PR upping the XCSoar version on master, if master-next is not ready yet.

mihu-ov:
I would appreciate it, and probably many other OV users as well!

linuxianer99:
Hi All :-)
Think that should be possible to build an image .. but ...

There are some breaking changes:

- Versioning is not longer <year, dateofbuild> but just a number based on Yocto versioning. So i have no clue, how ovmenu-ng, the update stick, etc are handling this.

Proposal: 
I can build an "release" image, and somebody did some tests on this ?
I would not say it is an official release .. mabye we break some Openvarios doing so ...

What so you think ?

mihu-ov:
Hi Timo,

This would be great! I could test this weekend but am away all next week and also the week after.
Is there a way to get the driver for Becker radios included?

@MaxKellermann 
Do you still use the Openvario I sent you? If yes, that´s great! If you don´t use it anymore, would you be able to send it back to me? My OV is installed in my glider and once the season starts it´s getting hard to test new images.

linuxianer99:
@mihu-ov : What is this driver about ?
Is not just a serial port ??

mihu-ov:
On the radio side, it is a RS422 serial port. In XCSoar there was a pull request for implementation of a driver for Becker radios that was rejected, as far as I remember mostly because the documentation is not made public by Becker. The driver is included in @August2111 OpenSoar releases so he can probably fill in the details.

mihu-ov:
I am happy to see a lot of updates in the last days! Are there scathgap images available already, or can anybody provide instructions how to build (for PQ-070 in my case)? Or is it still too early to test?

linuxianer99:
For the OV-CB2 i can build some images.
On my SteFly it works well, but it is CH070. 

Only untested feature up to now is software flash. As i am using Opensoar, it is a bit different.

@lordfolken @Maxel01 @mihu-ov Should we make a PRE-1 Image ??

mihu-ov:
I think yes, we need a PRE-1 image for testing to find issues before the gliding season is in full progress.

Maxel01:
[...]

PR #431, opened 2026-04-10
Rework of system init and addition of app selector

Scumi:
- use systemd and init script for startup
- allow both application being installed the same time
- give user option to chose app on first boot and when going into the shell
- ~add 3rd partition (/home instead of /data) on build, allow user to put his files there before first boot~
- ~expand 3rd partition on first boot~
- use `ovmenu-ng.sh` as entry point for either applications menu

As far as I see during my tests, both application work properly.


Scumi:
Can touchscreen-only users interact with the terminal dialog? If not, I suppose there should be a timeout for the inital app selection.

linuxianer99 (approved):
Looks fine for me :-)

linuxianer99:
@Maxel01 Do you want to do a review ??

Maxel01:
I don't have time for that in the next 2 weeks. So you can merge it if you Anprobe the changes.

Scumi:
Just not sure about the touchscreen thingy. Should I add a 60 sec counter on first boot?

linuxianer99:
> Just not sure about the touchscreen thingy. Should I add a 60 sec counter on first boot?

Fine for me :-) so we are 100% sure, that everybody gets a running system.

Scumi:
Okay. Will update the PR tomorrow probably.

Scumi:
Considering https://github.com/Openvario/meta-openvario/issues/391#issuecomment-4239526699, I would revert the image layout changes. Do you agree? Needs some work so XCSoar finds its data path.

August2111:
The 'app selector' should be part of the OpenVario menu! This should change the 'main_app' in the config.uEnv in boot partition. After selection, an immediate restart is required, which should be offered directly from the menu. 
The data folder for XCSoar is 'data/XCSoarData' – located right next to 'data/OpenSoarData' – and should also contain the same data as the original '.xcsoar' folder.

August2111:
In my opinion, it would be much more important to integrate the OpenVario menu into XCSoar (of course for OpenVario only) as soon as possible – then switching between the 'main application', firmware updates, Wi-Fi settings, and all other settings could be done transparently in both applications – and you wouldn't actually need to do anything in the shell...

Scumi:
> The 'app selector' should be part of the OpenVario menu! This should change the 'main_app' in the config.uEnv in boot partition. After selection, an immediate restart is required, which should be offered directly from the menu.

If I understand your concern correctly, the current implementation already behaves as described (see commit 6881e4a454d77d767893b3fff49eb22c3d77144a), with the only difference being that no explicit restart is triggered.

From my perspective, a restart is not required, since the selected application (including correct "menu") is started immediately after updating `config.uEnv`.

The app selector is also intentionally not shown on every startup. It only appears on first boot or after entering the shell, to avoid unnecessarily interrupting the user. After the initial selection, OpenSoar behaves exactly as before.

Please have a look at the commit contents and messages - the behavior should be documented there.

> In my opinion, it would be much more important to integrate the OpenVario menu into XCSoar (of course for OpenVario only) as soon as possible – then switching between the 'main application', firmware updates, Wi-Fi settings, and all other settings could be done transparently in both applications – and you wouldn't actually need to do anything in the shell...

This is already work in progress, but it will take some time and coordination (as you also pointed out). For reference:
- https://github.com/XCSoar/XCSoar/pull/2158
- https://github.com/Openvario/meta-openvario/discussions/437

mihu-ov:
[https://github.com/XCSoar/XCSoar/pull/2158](https://github.com/XCSoar/XCSoar/pull/2158l) looks great! I would love to see it in Openvario before the northern hemisphere gliding season is in full progress.

Scumi:
Included the app-selection including timer on first boot and took out the commits regarding partition layout. That topic needs refinement (or no action at all, lets see).

linuxianer99:
@Scumi Great Job!! 
So i will merge and trigger a release ??

Scumi:
Yes, from my point of view this is done. I fly with this config now. Need a spare OpenVario to continue development (networking should be next).

PR #441, opened 2026-05-27
Fix/cleanup

Scumi:
Some clean-ups and de-duplications. 
The WiFi change for RPi4 is for the upcoming xcsoar 7.45, which includes WiFi management in-app.
The psplash handover issues appear to be fixed as well (at least during my tests).


linuxianer99 (dismissed):

linuxianer99 (approved):

conf/distro/ovlinux.conf
```diff
@@ -93,7 +98,6 @@ DISTRO_FEATURES ?= "${DISTRO_FEATURES_DEFAULT} ${DISTRO_FEATURES_LIBC} ${POKY_DE
 DISTRO_FEATURES:append = " systemd usrmerge"
 DISTRO_FEATURES:remove = "acl argp pcmcia usbgadget xattr nfs zeroconf pci 3g nfc x11 sysvinit"
 DISTRO_FEATURES:remove:cubieboard2 = "gobject-introspection-data"
```
Maxel01: I needed the cubieboard overwrite in #409 when I added a dependency (I don't know which anymore but with the deletion of the following lines the overwrite can be removed too, maybe the python dependency too????). Maybe you could take a look at that PR.

Maxel01 (changes requested):
LGTM, I can't test the build right now. @Scumi you have a CM4 now right?

conf/distro/ovlinux.conf
Scumi: I'll check. 👍

Maxel01 (approved):

conf/distro/ovlinux.conf
Scumi: I re-added it, but it could stay removed. Its a tiny anyway, so I'd not optimize on that end further for now :-)

Maxel01: Yes, but for both hardwares.

PR #447, opened 2026-06-06
RepoMgmnt: Better structure for PRs

linuxianer99:
Add a template for PRs

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,45 @@
+Thank you for contributing to Openvario!
+
+We truly appreciate your time and effort in making Openvario better. We know
+that contributing to an open-source project can be challenging, and we're
+grateful that you've chosen to help.
+
+This checklist is here to help guide you through the submission process.
+Don't worry if you're not sure about something—feel free to ask questions
+or submit your PR, and we'll work together to get it ready.
+
+## Pre-Submission Checklist
+
+Please verify the following before submitting your PR:
+
+### Git & Commit History
+
+#### Branch & Rebase
+- [ ] PR is rebased on current branch you wanna commit to 
+- [ ] Use `git rebase -i` to clean up commit history before PR
+  submission
+
+#### Commit Format & Messages
+- [ ] All commits follow the format: `<Component>: <Summary>` (no
+  `src/` prefix)
+- [ ] Use present tense in commit messages ("Fix" not "Fixed", "Add"
+  not "Added")
+- [ ] Commit messages explain *why* the change was made, not just
+  *what* changed
+- [ ] Commit message body (if needed) provides detailed reasoning and
+  context
+
+#### Commit Structure
+- [ ] Each commit is atomic and builds successfully (every commit
+  must compile)
+- [ ] One commit per logical change (don't mix refactoring with
+  feature changes)
+- [ ] Self-contained commits (each commit changes one thing)
+- [ ] No fixup commits (squashed into parent commits using
+  `git rebase -i`)
+- [ ] No duplicate commits (check with `git log --oneline`)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,45 @@
+Thank you for contributing to Openvario!
+
+We truly appreciate your time and effort in making Openvario better. We know
+that contributing to an open-source project can be challenging, and we're
+grateful that you've chosen to help.
+
+This checklist is here to help guide you through the submission process.
+Don't worry if you're not sure about something—feel free to ask questions
+or submit your PR, and we'll work together to get it ready.
```
Maxel01: This could be comment or should it be visible in the PR in the end? (`<!-- -->`)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,45 @@
...
+  not "Added")
+- [ ] Commit messages explain *why* the change was made, not just
+  *what* changed
+- [ ] Commit message body (if needed) provides detailed reasoning and
+  context
+
+#### Commit Structure
+- [ ] Each commit is atomic and builds successfully (every commit
+  must compile)
+- [ ] One commit per logical change (don't mix refactoring with
```
Maxel01, line 35: That's correct, but how will you verify that?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,45 @@
...
+## Pre-Submission Checklist
+
+Please verify the following before submitting your PR:
+
+### Git & Commit History
+
+#### Branch & Rebase
+- [ ] PR is rebased on current branch you wanna commit to 
+- [ ] Use `git rebase -i` to clean up commit history before PR
+  submission
```
Maxel01, line 20: There are many new lines. Did you want to have it like that?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Commit Format & Messages
```
Maxel01, line 22: That's something I already mentioned earlier. There are multiple ways to handle PRs. You can merge all the commits which means that the contributors decide how the commits are named. The alternative is that a PR already meet these requirements as they would be easier to review. You could then squash several commits into a single one.

Maxel01 (approved):
Hi,
thank you for creating a template, this will definitely improve the PR workflow.
However at the moment it is just a checklist.
I added some comments. I just like the way how Nuxt is handling PRs and the automatic changelog generation. Just create a PR for one single purpose, create as many commits as you like. In the end it gets squashed into a single commit.
**But** nuxt is a full stack framework and has a totally different structure. Maybe you can't isolate all the changes in this case.
For me it's fine, just some ideas and how I personally handle PRs.
I could also create some templates for the issues to create a better structure.

.github/PULL_REQUEST_TEMPLATE.md
linuxianer99: is just comment ... should not be visible ...

.github/PULL_REQUEST_TEMPLATE.md
linuxianer99: Think i can not verify ... but is should just be a guideline

.github/PULL_REQUEST_TEMPLATE.md
linuxianer99: no :-)

.github/PULL_REQUEST_TEMPLATE.md
linuxianer99: I guess a PR can have commit in multiple components of the image ... So i make sense, to have the single commits in a structured way. The whole PR can be the complete feature ...

PR #458, opened 2026-09-23
xcsoar: update to 7.45.3

Scumi:


lordfolken (approved):

linuxianer99 (approved):

linuxianer99:
Now it would be great, if there would be a possibility to just update XCSoar in the OV Image via a package manager ..
What do you think ??

Scumi:
Well, either that, or use RAUC bundled updates. I'm steadily work towards the latter, which would use either OTA or USB / SD offline updates. But will take the winter until its ready I guess.

